PR #4, opened 2015-01-23
feat(formatters): add autopep8

actionless:
mb it make sense to generalize them somehow?
because it's almost the same as gofmt one

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -182,7 +182,7 @@ if !exists('s:autopep8')
     let l:executable = s:plugin.Flag('autopep8_executable')
     if !exists('s:autopep8_supports_range')
       let l:version_call =
-          \ maktaba#syscall#Create([l:executable, '--version', '1>&2']).Call()
+          \ maktaba#syscall#Create([l:executable, '--version']).Call()
       let l:version_output =
           \version_call.stderr ? version_call.stderr : version_call.stdout
       let s:autopep8_supports_range =
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -194,8 +194,9 @@ if !exists('s:autopep8')
     let l:lines = getline(1, line('$'))
 
     if s:autopep8_supports_range
-      let l:cmd = [ l:executable, '-', '--range',
-                  \ ''.a:startline, ''.a:endline ]
+      let l:cmd = [ l:executable,
+                  \ '--range', ''.a:startline, ''.a:endline,
+                  \ '-' ]
       let l:input = join(l:lines, "\n")
     else
       let l:cmd = [ l:executable, '-' ]
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -179,18 +179,38 @@ if !exists('s:autopep8')
   " @throws ShellError
   function s:autopep8.FormatRange(startline, endline) abort
     " Hack range formatting by formatting range individually, ignoring context.
-    let l:cmd = [ s:plugin.Flag('autopep8_executable'), "-"  ]
+    let l:executable = s:plugin.Flag('autopep8_executable')
+    if !exists('s:autopep8_supports_range')
+      let l:version_output =
+          \ maktaba#syscall#Create([l:executable, '--version']).Call().stderr
+      let s:autopep8_supports_range =
+          \ matchlist(l:version_output, '\m\Cautopep8 \(\d\+\)\.')[1] >= 1
+    endif
+
     call maktaba#ensure#IsNumber(a:startline)
     call maktaba#ensure#IsNumber(a:endline)
     let l:lines = getline(1, line('$'))
-    let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
+
+    if s:autopep8_supports_range
+      let l:cmd = [ l:executable, '-', '--range',
+                  \ ''.a:startline, ''.a:endline ]
+      let l:input = join(l:lines, "\n")
+    else
+      let l:cmd = [ l:executable, '-' ]
+      let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
+    endif
 
     let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
     let l:formatted = split(l:result.stdout, "\n")
-    " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
-    let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
 
-    let l:full_formatted = l:before + l:formatted + l:lines[a:endline :]
+    if s:autopep8_supports_range
+      let l:full_formatted = l:formatted
+    else
+      " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
+      let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
+      let l:full_formatted = l:before + l:formatted + l:lines[a:endline :]
+    endif
+
     call maktaba#buffer#Overwrite(1, line('$'), l:full_formatted)
   endfunction
 
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -181,8 +181,10 @@ if !exists('s:autopep8')
     " Hack range formatting by formatting range individually, ignoring context.
     let l:executable = s:plugin.Flag('autopep8_executable')
     if !exists('s:autopep8_supports_range')
+      let l:version_call =
+          \ maktaba#syscall#Create([l:executable, '--version', '1>&2']).Call()
       let l:version_output =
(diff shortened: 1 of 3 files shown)


autoload/codefmt.vim
```diff
@@ -156,7 +156,67 @@ if !exists('s:gofmt')
   call codefmtlib#AddDefaultFormatter(s:gofmt)
 endif
 
+" Formatter: autopep8
+if !exists('s:autopep8')
+  let s:autopep8 = {
+      \ 'name': 'autopep8',
+      \ 'setup_instructions': 'Install autopep8 and '}
```
fowles: the setup instructions seem to trail off

fowles:
It might make sense to generalize now that there is more than one. 

Before merging any request you have we will need you to sign the CLA (contributor lisence agreement).  Sorry the lawyercats are sticklers for this.

https://github.com/google/vim-codefmt/blob/master/CONTRIBUTING.md

autoload/codefmt.vim
```diff
@@ -156,7 +156,67 @@ if !exists('s:gofmt')
...
+  " Reformat the current buffer with autopep8 or the binary named in
+  " @flag(autopep8_executable), only targeting the range between {startline} and
+  " {endline}.
+  function s:autopep8.FormatRange(startline, endline) abort
+    " Hack range formatting by formatting range individually, ignoring context.
+    let l:cmd = [ s:plugin.Flag('autopep8_executable'), "-"  ]
+    call maktaba#ensure#IsNumber(a:startline)
+    call maktaba#ensure#IsNumber(a:endline)
+    let l:lines = getline(1, line('$'))
+    let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
```
dbarnett: autopep8 1.0 and later support `--range line line`, which will give much better results than this hack. Unfortunately, Ubuntu seems to be stuck on a version almost 2 years old (0.9.1) which doesn't support it, so I could see people being stuck on an old version without it. I just filed https://bugs.launchpad.net/ubuntu/+source/autopep8/+bug/1414258 asking Ubuntu to update.

Could you file a feature request along with these changes to support true range formatting, with this as a fallback? "autopep8 --version" runs pretty quick for me, so I don't think running that once per vim session to detect native range support would be a problem.

dbarnett:
There's a list of formatters on line 28 that should be updated to include python, and then if you can run [vimdoc](google/vimdoc) on it after that it will update the help file in doc/ to match.

autoload/codefmt.vim
actionless: how do u think it will be better to make that choice -- using plugin.Flag or reading "autopep --version" output once and caching it?

dbarnett: If reading `autopep8 --version` works well, I'd go with that approach. Seems like it will.

actionless: btw, while i gonna store cached value in Flag, then it will be possible to set it manually

or Flags are not suitable for such purpose and intended to be _only_ for storing some default settings, but not any variable data?

dbarnett: Using a flag like that wouldn't be unreasonable, except I figured it should be lazy and flag initialization should be eager (once at plugin installation time). Even though it's a quick check, doing it for N formatters plus the `executable()` checks all at once might not be.

actionless: i mean to add that version check not on plugin initialization but inside `FormatRange` function, like:
```vim
if !s:plugin.Flag('autopep8_version')
    s:plugin.Flag('autopep8_version', GetVersion())
```

dbarnett: Yeah, I thought lazy flag initialization like that might get confusing, but maybe not. Anyway, my vote would be to wait to add a flag until someone reports they need to override the behavior.

actionless: am i understood right, you mean to initialize version flag like that but not to set it manually by default?

autoload/codefmt.vim
```diff
@@ -156,7 +157,68 @@ if !exists('s:gofmt')
...
+    try
+      let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
+      let l:formatted = split(l:result.stdout, "\n")
+      " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
+      let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
+
+      let l:full_formatted = l:before + l:formatted + l:lines[a:endline :]
```
actionless: oops, just noticed what autopep8 don't output errors like that.

actually i am not even sure if it can return unsuccessful result when using stdin at all (when it faces wrong syntax or smth it is just ignoring it)

autoload/codefmt.vim
dbarnett: What I was suggesting is instead of using a flag, for now use a script-local variable like
```VimL
let l:executable = s:plugin.Flag('autopep8_executable')
if !exists('s:autopep8_supports_range')
  let l:version_output =
      \ maktaba#syscall#Create([l:executable, '--version']).Call().stderr
  let s:autopep8_supports_range =
      \ matchlist(l:version_output, '\m\Cautopep8 \(\d\+\)\.')[1] >= 1
endif
```
Then, if for some reason a user reports the version check is not right for them and they'd rather configure it manually, it's easy to replace the script-local variable with flag operations.

autoload/codefmt.vim
dbarnett: Yeah, it's fine to get rid of the try/catch and just let the ShellError bubble up. It's caught and handled in codefmt. You can add a `@throws ShellError` line into the vimdoc for good measure.

autoload/codefmt.vim
actionless: got it, thanks

actionless:
i was playing with the `--range` option of autopep8 and noticed some problem -- it can affect the number of lines in the resulting file, so if file changed number of lines after autoformatting it's impossible to slice the initial range of lines from autopep8 output (and i did not found out how to output only range instead of whole file)

actionless:
i think it's ready for review

autoload/codefmt.vim
```diff
@@ -156,6 +157,65 @@ if !exists('s:gofmt')
...
+          \ maktaba#syscall#Create([l:executable, '--version']).Call().stderr
+      let s:autopep8_supports_range =
+          \ matchlist(l:version_output, '\m\Cautopep8 \(\d\+\)\.')[1] >= 1
+    endif
+
+    call maktaba#ensure#IsNumber(a:startline)
+    call maktaba#ensure#IsNumber(a:endline)
+    let l:lines = getline(1, line('$'))
+
+    if s:autopep8_supports_range
```
fowles: Can you add a comment that the '-' argument means stdin and put the '-' at the end of the list of args (I got a bit confused by the '-', '--range' list)

fowles:
I am going to wait for dbarnett's LGTM because he knows vimscripts subtleties bettter than I do.  

I see that you did sign the CLA, thanks for doing that!  The good news is that the CLA is good for all google stuff (except android), so you won't ever have to bother with that again.

autoload/codefmt.vim
dbarnett: I didn't find it too confusing, but it could be more obvious at a glance if the lines were broken up differently:
```VimL
      let l:cmd = [
          \ l:executable,
          \ '--range', ''.a:startline, ''.a:endline,
          \ '-' ]
```

Matt, you still think a comment would be necessary like that? Was it more that you didn't notice the '-' or just didn't know what to make of it?

dbarnett:
LGTM. You can merge when you're happy with the `l:cmd` part.

autoload/codefmt.vim
fowles: Didn't know what to make of it. I thought it was similar to the bash things
which say "all arguments after this one are files" which confused me and
sent me off to read the docs.

Matt

autoload/codefmt.vim
```diff
@@ -156,6 +157,68 @@ if !exists('s:gofmt')
...
+  " {endline}.
+  " @throws ShellError
+  function s:autopep8.FormatRange(startline, endline) abort
+    " Hack range formatting by formatting range individually, ignoring context.
+    let l:executable = s:plugin.Flag('autopep8_executable')
+    if !exists('s:autopep8_supports_range')
+      let l:version_call =
+          \ maktaba#syscall#Create([l:executable, '--version']).Call()
+      let l:version_output =
+          \version_call.stderr ? version_call.stderr : version_call.stdout
```
dbarnett: Can you add a comment on what cases you're handling here? To my eye the logic looks a little weird without any info about expected output.

BTW, it reads a little weird not having a space after the `\`.

actionless: so as i described in commit message, in some configuration version is being output into stdout, in some to stderr.
i have no clue why that happens (it seems what i have the same library versions on both PCs there i tested it,  but maybe some different environment variables or smth)

yup, that's a typo, i will fix it

fowles:
Thanks for the great patch!

PR #8, opened 2015-01-24
Fix vroom tests to not depend on executables installed on host

### LINKED ISSUE #7: vroom tests fail if clang-format and gofmt aren't installed on host
Even though the tests stub out the actual system calls to clang-format and gofmt, they're not completely hermetic because they depend on whether 'clang-format' and 'gofmt' are actual executables on the path on the host where the tests are being run.

This causes vroom failures like
```
FAILED on line 030: Multiple failures:
Unexpected message:
Not available. codefmt doesn't have a default formatter for this buffer.

Expected system call not received.

Got no chance to inject response: 
 EXPECT:	clang-format -style file -assume-filename .*foo.cc .*2>.* (regex mode)
 STDOUT:	void f() {
       	  int i;
       	  SomeFunction(parameter,  // comment
       	               i);
       	}

Messages:

Not available. codefmt doesn't have a default formatter for this buffer.

Failed command on line 030:
:FormatCode

Queued system controls are:
 EXPECT:	clang-format -style file -assume-filename .*foo.cc .*2>.* (regex mode)
 STDOUT:	void f() {
       	  int i;
       	  SomeFunction(parameter,  // comment
       	               i);
       	}

No system calls received. Perhaps your --shell is broken?

Last few commands (most recent last) were:
:let g:repo = fnamemodify($VROOMFILE, ':p:h:h')<CR>
:execute 'source' g:repo . '/bootstrap.vim'<CR>
:call maktaba#syscall#SetUsableShellRegex('\v<shell\.vroomfaker$')<CR>
:filetype plugin on<CR>
:let g:repeat_calls = []<CR>
:function FakeRepeat(...)<CR>  call add(g:repeat_calls, a:000)<CR><CR>
:endfunction<CR>
:call maktaba#test#Override('repeat#set', 'FakeRepeat')<CR>
:silent file foo.cc<CR>
:set filetype=cpp<CR>
ivoid f() {int i; SomeFunction(parameter,// comment<CR>i);}<ESC>
:FormatCode<CR>
```

There's no way to actually stub out `executable()` calls in vroom, so we'll either need to provide some override to actually change the plugin behavior for testing or monkey patch the `IsAvailable()` function on the formatters we're testing.

dbarnett:
Fixes #7.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -158,10 +158,22 @@ endif
 
 
 ""
+" Checks whether {formatter} is available.
+" NOTE: If @function(#DisableIsAvailableChecksForTesting) has been called, skips
+" the IsAvailable check and always returns true.
+function! s:IsAvailable(formatter) abort
+  if get(s:, 'check_formatters_available', 1)
+    return a:formatter.IsAvailable()
+  endif
+  return 1
+endfunction
+
+
+""
 " Detects whether a formatter has been defined for the current buffer/filetype.
 function! codefmt#IsFormatterAvailable() abort
   let l:formatters = copy(codefmtlib#GetFormatters())
-  let l:is_available = 'v:val.AppliesToBuffer() && v:val.IsAvailable()'
+  let l:is_available = 'v:val.AppliesToBuffer() && s:IsAvailable(v:val)'
   return !empty(filter(l:formatters, l:is_available)) ||
         \ !empty(get(b:, 'codefmt_formatter'))
 endfunction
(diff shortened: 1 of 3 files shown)


dbarnett:
My first attempt (2f55a1790f) kept the mess contained in the vroom files, but was really, really nasty. I could go either way, but I'm leaning towards the codefmt#DisableIsAvailableChecksForTesting helper approach.

malcolmr:
LGTM. 3d90013 is much nicer.

Issue #11, opened 2015-01-24
Clean up codefmt vroom files


dbarnett:
The codefmt vroom files are pretty good for test coverage right now but not very friendly and readable as "executable documentation". Compare to the glaive docs at https://github.com/google/vim-glaive/tree/master/vroom as an A+ example. Improvements to consider:
  * Add more introductory text and a "table of contents" in main.vroom. Consider renaming to codefmt.vroom.
  * Split off a separate formatters.vroom file for testing the built-in formatters and consider using very simple dummy formatters in the other files.
  * Cut down on repetition. It's not easy right now to see at a glance what each executable section is testing/demonstrating for all of the verbose syntax.

PR #17, opened 2015-01-26
Improve error message for unavailable plugins.

### LINKED ISSUE #16: Error: Not available. codefmt doesn't have a default formatter for this buffer.
[The announcement](http://google-opensource.blogspot.ru/2015/01/format-code-easy-way-with-codefmt.html) claims this plugin supports cpp & js natively.
so after installing with vundle ( as provided in the readme file )
when I try to `:FormatCode` on .cpp or .js files I get following error:
`Not available. codefmt doesn't have a default formatter for this buffer.`

What I am missing? Should one use VAM or install any additional plugins? Thanks in advance.

fowles:
Fixes #16

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -222,19 +222,17 @@ function! s:GetFormatter(...) abort
       " Report a better error message in that case.
       let l:unavailable_formatters = filter(
             \ copy(l:formatters), 'v:val.AppliesToBuffer()')
+      let l:error = ''
       if !empty(l:unavailable_formatters)
-        let l:error = ''
         for l:formatter in l:unavailable_formatters
           let l:error .= GetSetupInstructions(l:formatter) . '\n'
         endfor
-        call maktaba#error#Shout(l:error)
-        return
       else
-        call maktaba#error#Shout(
-              \ 'Not available. codefmt doesn''t have a default formatter for ' .
-              \ 'this buffer.')
-        return
+        let l:error = 'Not available. codefmt doesn''t have a default ' .
+              \ 'formatter for this buffer.'
       endif
+      call maktaba#error#Shout(l:error)
+      return
     endif
   endif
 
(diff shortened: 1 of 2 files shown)


autoload/codefmt.vim
```diff
@@ -279,9 +282,23 @@ function! s:GetFormatter(...) abort
...
       let l:formatter = l:default_formatters[0]
     else
-      call maktaba#error#Shout(
-          \ 'Not available. codefmt doesn''t have a default formatter for ' .
-          \ 'this buffer.')
+      " Check if we have formatters that are not avialable for some reason.
+      " Report a better error message in that case.
+      let l:unavailable_formatters = filter(
+            \ copy(l:formatters), 'v:val.AppliesToBuffer()')
+      let l:error = ''
```
dbarnett: You can replace this whole thing with `let l:error = join(map(copy(l:unavailable_formatters), 's:GetSetupInstructions(v:val)'), "\n")`. Doesn't have to be all one line, but using `join()` would be a big improvement, and `map()` probably would too.

vroom/main.vroom
```diff
@@ -18,7 +18,30 @@ stub that out for vroom.
   :endfunction
   :call maktaba#test#Override('repeat#set', 'FakeRepeat')
 
-We'll also stub it out to not care whether certain executables are installed on
+
+Before we stub out availability checks, test :FormatCode for unavailable plugins
```
dbarnett: Hmm, putting this gnarliness at the very beginning of the file would undermine the sequence of concepts I'm trying to build in #11. Could we just add a `codefmt#ReenableIsAvailableChecksForTesting()` or `codefmt#OverrideIsAvailableChecksForTesting({is_enabled})` and put further down in the file where it's more natural?

fowles: Actually, I have spent the last hour or so trying to make that work.  I clearly don't understand something about vim's evaluation of globals or something.  Do you mind taking a crack at it?

dbarnett: Heh, I know how you feel. Sure, if you want to just address the other comment and submit, I can take care of it.

dbarnett:
LGTM modulo the loop cleanup.

PR #21, opened 2015-01-27
Clean up vroom files to introduce concepts better, add autopep8 coverage

### LINKED ISSUE #11: Clean up codefmt vroom files

dbarnett:
Cleans up vroom files so they explain what they're doing more clearly and introduce concepts in a more natural order. Cuts down on a lot of repetition by leaving out details that don't relate to the feature being demonstrated/tested, like using more wildcards in syscall lines and formatting empty buffers in places.

Also adds a lot of coverage and adds testing for autopep8 which previously had none.

Fixes #11.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -184,13 +184,13 @@ if !exists('s:autopep8')
           \ maktaba#syscall#Create([l:executable, '--version']).Call()
       " In some cases version is written to stderr, in some to stdout
       let l:version_output = empty(version_call.stderr) ?
-            \ version_call.stdout : version_call.stderr
+          \ version_call.stdout : version_call.stderr
       let l:autopep8_version =
           \ matchlist(l:version_output, '\m\Cautopep8 \(\d\+\)\.')
       if empty(l:autopep8_version)
         throw maktaba#error#Failure(
-              \ 'Unable to parse version from `%s --version`: %s',
-              \ l:executable, l:version_output)
+            \ 'Unable to parse version from `%s --version`: %s',
+            \ l:executable, l:version_output)
       else
         let s:autopep8_supports_range = l:autopep8_version[1] >= 1
       endif
@@ -289,7 +289,7 @@ function! s:GetFormatter(...) abort
     if !empty(l:default_formatters)
       let l:formatter = l:default_formatters[0]
     else
-      " Check if we have formatters that are not avialable for some reason.
+      " Check if we have formatters that are not available for some reason.
       " Report a better error message in that case.
       let l:unavailable_formatters = filter(
             \ copy(l:formatters), 'v:val.AppliesToBuffer()')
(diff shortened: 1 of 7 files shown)


malcolmr:
Looks great; CI tests are failing, though?

autoload/codefmt.vim
```diff
@@ -393,3 +393,21 @@ endfunction
...
 endfunction
+
+
+""
+" @private
+" Undoes the effect of @function(#DisableIsAvailableChecksForTesting), allowing
+" FORMATTER.IsAvailable checks to work normally.
+function! codefmt#DontDisableIsAvailableChecksForTesting() abort
+  let s:check_formatters_available = 1
+endfunction
```
fowles: Seems like one method that takes an argument would be more natural.  If you want to stick with two methods, you should avoid double negatives.  DontDisable is better written Enable

vroom/autopep8.vroom
```diff
@@ -0,0 +1,115 @@
...
+  $ 
+
+The name or path of the autopep8 executable can be configured via the
+autopep8_executable flag if the default of "autopep8" doesn't work.
+
+  :Glaive codefmt autopep8_executable='autopep20'
+  :FormatCode autopep8
+  ! autopep20 .*
+  $ 
+  :Glaive codefmt autopep8_executable='autopep8'
```
fowles: Are these glaive commands clearing the cached --version?  Should we use that everywhere?

dbarnett: Good idea. Then I can get rid of `codefmt#ForgetAutopep8VersionForTesting` and just overwrite the flag instead.

autoload/codefmt.vim
dbarnett: Any suggestions for a name for the single method? `codefmt#OverrideWhetherToDoIsAvailableChecksForTesting({enable})`? Long names should be fine for this as long as they're clear at a glance.

fowles: codefmt#SetWhetherToPerformIsAvailableChecksForTesting

dbarnett:
I tracked down the issue with the tests to google/vroom#84. Pretty gnarly issue. Worked around it here by adding placeholder content everywhere.

Also updated the autopep8 version caching and testing hooks as suggested. PTAL.

PR #22, opened 2015-01-27
feat(formatters: js-beautify): add formatter for js/css/html/json

actionless:


instant/flags.vim
```diff
@@ -51,3 +51,11 @@ call s:plugin.Flag('clang_format_style', 'file')
...
 " additionally adjust imports when formatting.
 call s:plugin.Flag('gofmt_executable', 'gofmt')
+
+""
+" The path to the js-beautify executable.
+call s:plugin.Flag('js_beautify_executable', 'js-beautify')
+
+""
+" Default formatter for javascript files, 'clang_format' or 'js_beautify'.
+call s:plugin.Flag('javascript_formatter', 'clang_format')
```
fowles: I am torn about this flag.  If you have js-beautify installed but not clang-format installed, it would be nice if we automatically picked up the one that worked.  I don't know a good solution though.  I am somewhat tempted to just leave them both installed and ask people to override the default with b:codefmt_buffer and some kind of hook...

Anyone else have thoughts?

Can you merge https://github.com/google/vim-codefmt/pull/21 and add tests for js-beautify too?

actionless: mb we can do smth like that?
```viml
call s:plugin.Flag(
  'javascript_formatter',
  executable(s:plugin.Flag('clang_format_executable')) ? 'clang_format' : 'js_beautify'
)
```


and about tests, i will wait until #21 will pass (to have a proper working example) and then will add the tests to this PR

fowles: That solves the immediate problem, but I wonder if there is something more general we should be thinking about.  We are going to get more formatters that overlap like this (we have multiple java formatters at google).  Putting a new flag for every one of them would work, but doesn't seem like the best solution.  I am trying to think if we can generalize in some fashion...

autoload/codefmt.vim
```diff
@@ -228,6 +229,54 @@ if !exists('s:autopep8')
   call codefmtlib#AddDefaultFormatter(s:autopep8)
 endif
 
+
+" Formatter: js-beautify
+if !exists('s:js_beautify')
+  let s:js_beautify = {
+      \ 'name': 'js_beautify',
```
dbarnett: I'd use a dash here to match the executable name, like we do with clang-format.

autoload/codefmt.vim
```diff
@@ -228,6 +229,54 @@ if !exists('s:autopep8')
...
+  endfunction
+
+  ""
+  " Reformat the current buffer with js-beautify or the binary named in
+  " @flag(js_beautify_executable), only targeting the range between {startline} and
+  " {endline}.
+  " @throws ShellError
+  function s:js_beautify.FormatRange(startline, endline) abort
+    let l:cmd = [ s:plugin.Flag('js_beautify_executable'),
+                \'--type', &filetype,
```
dbarnett: Have you tested this when filetype is empty?

autoload/codefmt.vim
```diff
@@ -228,6 +229,54 @@ if !exists('s:autopep8')
...
+                \'-f', '-' ]
+
+    call maktaba#ensure#IsNumber(a:startline)
+    call maktaba#ensure#IsNumber(a:endline)
+
+    let l:lines = getline(1, line('$'))
+    " Hack range formatting by formatting range individually, ignoring context.
```
dbarnett: Could you file an issue with js-beautify for range formatting and link to it here? This being a hack, it would be good to know we've at least followed up with the project maintainers.

instant/flags.vim
dbarnett: I'm inclined to say we don't need this flag. The default is just the default. If someone has a specific preference they can configure it using `b:codefmt_formatter` and autocmds. I would like to add some recipes for that into the codefmt docs, though.

There is one point worth mentioning: Registration order matters for formatters. As-is, clang-format will take precedence since it's registered before js-beautifier, but my guess is it shouldn't since it doesn't specialize in JS while js-beautify does.

I will probably want a blacklist flag to say "I don't have a specific formatter I want, but formatter X is annoying and I know I never want it by default".

autoload/codefmt.vim
actionless: i did not thought about the case of calling it with specifying formatter manually, i will fix it

instant/flags.vim
actionless: about lists, yesterday before the sleep i was thinking about it, but in my mind it was more like whitelist, not blacklist, ie to have it described like `l:preferred_formatters_js= [ 'js_beautify', 'clang_format' ]`(btw is it possible to store list as a flag?) and logic of codefmt#IsFormatterAvailable() to select most preferred formatter of all available ones

autoload/codefmt.vim
actionless: created: https://github.com/beautify-web/js-beautify/issues/610

fowles:
#21 merged into mainline.

actionless:
thanks for the notification, i will start playing with the tests soon

actionless:
btw does vroom have sorta cheatsheet for each specific symbol (like $, &)?

i was looking on examples but because those symbols are not phonetic it's too hard to remember them and takes a lot of time to search for it in examples each time when you forgot what one means

dbarnett:
> btw does vroom have sorta cheatsheet for each specific symbol (like $, &)?

If we rendered the examples to HTML or similar and published them on google.github.io/vroom, would that help make them more convenient to look up? We could also put a table of symbols in the README or something.

fowles:
+1 on the table of symbols to meanings.  I usually just find something else
that does what I want and copy/paste

Matt

dbarnett:
As far as the preferred_formatters approach, I think that could already be expressed exactly as
  autocmd Filetype javascript let b:codefmt_formatter = 'js-beautify'
in which case I wouldn't want to add a new flag just to express the same thing a different way.

fowles:
I agree, but we should put that little bit of magic into the README so
people can easily find it.

actionless:
about vroom symbols:
yup, table would be awesome

about preferred formatters:
so mb logic of all "s:formatter.AppliesToBuffer()" can be declaratively described as list autocmds?

dbarnett:
Re declaratively-described AppliesToBuffer, how would that look and what are the advantages?

actionless:
smth like:

```viml
autocmd FileType javascript setFormatter('js-beautify')
autocmd FileType python     setFormatter('autopep8')
```

benefit -- not to have different ways to do the same thing, ie more generalization/readability

dbarnett:
That wouldn't be a complete replacement because there's a list of default formatters, not a single value. The first available out of the list is used (e.g., right now if you only have clang-format available that will be used for JS, and if you only have js-beautify available that will be used). And it's better if the default logic lives together with the formatter so people remember to implement it.

That's not to say we shouldn't rethink the defaulting part of the configuration, I'm just not convinced moving everything to separate autocmds will help.

actionless:
i think it can be like that then:
```viml
autocmd FileType javascript setFormatter('js-beautify', 'clang_format')
```

and make that function for setting formatter to accept multiarguments (or array? i am a newbie in vim) and use first available formatter from that list

dbarnett:
Maybe you could explain more specifically the improvement you're going for. The other case we've seen is special filenames needing a certain formatter,

actionless:
so the idea is instead of having methods in each formatter which checks if formatter is applicable or not to describe those relation between target and formatter explicitly in one place

about filenames, i think it can be done with autocmd as well, isn't it? and "priority" of those rules will depend on their order

fowles:
My suggestion for this change is:

remove `javascript_formatter` flag (users can configure it via autocmd `b:codefmt_formatter`)

that way when we decide on the official way to handle priorities or the like later, we don't have to go through a deprecation phase.

actionless:
removed the flag

btw, can you somehow ping your colleagues regarding vroom cheatsheet?

dbarnett:
Ping received. =)

Do you have urgent questions blocking what you're doing here? I can't guarantee any of us will get to it in the next week or so. BTW, we did get google/vroom#85 filed for it.

actionless:
so i subscribed for that ticket now

the thing what i can just copy tests from other formatter and substitute example code and they should work somehow but i don't want to do it like that but want to understand what i am writing there

i will try to tackle vroom examles once again, if other sort of doc will be not be written soon

actionless:
isn't that too early?

i thought tests are necessary

fowles:
You are correct, I jumped the gun.  Will you still be a good person and write the tests anyway?

actionless:
yup, i gonna to do that

just need to schedule an hour or so to read those examples

fowles:
the test monster is hungry!!

PR #28, opened 2015-02-21
test(formatters: js-beautify): cover basics

actionless:
but we still need to figure out what to do with default formatters for filetype if more than one is available


fowles:
Thanks for doing that!

PR #34, opened 2015-03-20
Use Maktaba's extension registry rather than codefmtlib.

### LINKED ISSUE #32: Switch from codefmtlib to maktaba#extension
Maktaba now has a built-in `maktaba#extension#` mechanism that's designed to replace codefmtlib and similar plugins. We should switch codefmt to use it and deprecate the codefmtlib repo.

malcolmr:
This change removes the codefmt->codefmtlib dependency, replacing it with the new extension registry added in Maktaba 1.9.0.

In detail:
- Replace codefmtlib references with `maktaba#extensions`.
- Bump the requirements to Maktaba 1.9.0.
- Inline the documentation and EnsureFormatter() method from codefmtlib.
- Inline (and slightly adjust) some of the tests from codefmtlib.
- Register jsbeautify before clang-format so that clang-format is still the default formatter for filetype javascript. (codefmtlib was first-registration wins, maktaba#extensions is last-registration wins.)

Fixes #32.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -30,11 +30,120 @@
 "   * python: autopep8
 
 
-call maktaba#library#Require('codefmtlib')
-
-
 let s:plugin = maktaba#plugin#Get('codefmt')
 let s:plugin_root = expand('<sfile>:p:h:h:h')
+let s:registry = s:plugin.GetExtensionRegistry()
+
+call s:registry.SetValidator('codefmt#EnsureFormatter')
+
+
+""
+" @dict Formatter
+" Interface for applying formatting to lines of code.  Formatters are
+" registered with codefmt using maktaba's standard extension registry:
+" >
+"   let l:codefmt_registry = maktaba#extension#GetRegistry('codefmt')
+"   call l:codefmt_registry.AddExtension(l:formatter)
+" <
+"
+" Formatters define these fields:
+"   * name (string): The formatter name that will be exposed to users.
+"   * setup_instructions (string, optional): A string explaining to users how to
+"     make the plugin available if not already available.
+" and these functions:
+"   * IsAvailable() -> boolean: Whether the formatter is fully functional with
+"     all dependencies available. Returns 0 only if setup_instructions have not
+"     been followed.
+"   * AppliesToBuffer() -> boolean: Whether the current buffer is of a type
+"     normally formatted by this formatter. Normally based on 'filetype', but
+"     could depend on buffer name or other properties.
+" and should implement at least one of the following functions:
+"   * Format(): Formats the current buffer directly.
+"   * FormatRange({startline}, {endline}): Formats the current buffer, focusing
+"     on the range of lines from {startline} to {endline}.
+"   * FormatRanges({ranges}): Formats the current buffer, focusing on the given
+"     ranges of lines. Each range should be a 2-item list of
+"     [startline,endline].
+" Formatters should implement the most specific format method that is supported.
+
+
+""
+" @private
+" Ensures that {formatter} is a valid formatter, and then prepares it for use by
+" codefmt.  See @dict(Formatter) for the API {formatter} must implement.
+" @throws BadValue if {formatter} is missing required fields.
+" Returns the fully prepared formatter.
+function! codefmt#EnsureFormatter(formatter) abort
+  let l:required_fields = ['name', 'IsAvailable', 'AppliesToBuffer']
+  " Throw BadValue if any required fields are missing.
+  let l:missing_fields =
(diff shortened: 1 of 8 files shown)


autoload/codefmt.vim
```diff
@@ -454,3 +516,17 @@ endfunction
 function! codefmt#SetWhetherToPerformIsAvailableChecksForTesting(enable) abort
   let s:check_formatters_available = a:enable
 endfunction
+
+""
+" @private
+" Returns all formatters, for debugging.
```
malcolmr: Not sure if we want to keep these two around. I found them essential for debugging the tests, though.

dbarnett: Tab completion on `:FormatCode` gets you the list of all formatters (with applicable formatters first). If we need more info for debugging, could we add a `:CodeFormatInfo` command to help with that and user troubleshooting?

autoload/codefmt.vim
```diff
@@ -30,11 +30,121 @@
...
+"     ranges of lines. Each range should be a 2-item list of
+"     [startline,endline].
+" Formatters should implement the most specific format method that is supported.
+
+
+""
+" @private
+" Ensures that {formatter} is a valid formatter, and then prepares it for use by
+" codefmt.  See @dict(Formatter) for the API {formatter} must implement.
+" @throws WrongType if {formatter} is not a dict.
```
dbarnett: Remove `@throws WrongType`. Apparently handled at a higher level now.

autoload/codefmt.vim
```diff
@@ -30,11 +30,121 @@
...
+          \ ' has no format functions.  It must have at least one of ' .
+          \ join(l:available_format_functions, ', '))
+  endif
+
+  " TODO(dbarnett): Check types.
+
+endfunction
+
+
+" Formatter: js-beautify
```
dbarnett: Any reason for this reordering?

vroom/main.vroom
```diff
@@ -185,8 +220,8 @@ them are available, the message will show a line for each.
   :let fake_format2 = copy(fake_format)
   :let fake_format2.name = 'fake-format2'
   :let fake_format2.setup_instructions = 'LMGTFY'
-  :call codefmtlib#AddFormatter(fake_format2)
+  :call maktaba#extension#GetRegistry('codefmt').AddExtension(fake_format2)
 
   :FormatCode
-  ~ Formatter "fake-format" is not available. Setup instructions: RTFM
   ~ Formatter "fake-format2" is not available. Setup instructions: LMGTFY
+  ~ Formatter "fake-format" is not available. Setup instructions: RTFM
```
dbarnett: What's up with the order here? Not that it's a huge deal, but I thought we preserved order for first-wins so I'm surprised it changed here.

dbarnett:
LGTM with minor comments.

autoload/codefmt.vim
malcolmr: Done. (I removed the `throw`, but forgot the `@throws`.)

autoload/codefmt.vim
malcolmr: See below.

autoload/codefmt.vim
malcolmr: Sounds good, and thanks for the hint about `:FormatCode`. Removed these.

vroom/main.vroom
malcolmr: codefmtlib did in-order registration (with two priorities), and then codefmt did first-wins over [external + internal].

First-wins is much more convenient for callers, so we retained that in `maktaba#extensions`, but in-order registration doesn't let you override an existing registration, so `maktaba#extensions` inserts at the _start_ of the respective lists.

That means that if you register formatter `foo` and later register `bar`, we now check `bar` first to give it a chance of overriding `foo`.  That normally doesn't matter, but in this case (and clang-format/jsbeautify above), it does, as both formatters handle the same filetypes.

malcolmr:
@dbarnett PTAL, mostly to check you're okay with the registration ordering change.

vroom/main.vroom
dbarnett: Oh, right, forgot we did change precedence. So to be clear, last registered wins and we're just using "first-wins" to describe the recommended handling of the list from GetFormatters, right? Still makes sense to me, it's just subtle enough I keep forgetting.

dbarnett:
LGTM.

vroom/main.vroom
malcolmr: That's right: callers should do first-match-wins, and to allow overrides (last-registered wins), we insert into the front of the list.

PR #36, opened 2015-03-20
Test against the version of Maktaba we require.

### LINKED ISSUE #35: codefmt CI tests should checkout a specific version of Maktaba
codefmt depends upon (and documents) a minimum version of Maktaba (currently 1.1; soon to be 1.9). The codefmt CI tests should check out that version, not HEAD, so that we can make sure that we can work with the version we've specified.

(Locally, codefmt will use whatever's available locally, which _will_ probably be HEAD, so I don't think we'd be losing any test coverage by doing that.)

Unless anyone objects, I'd like to:
- Push tags to `vim-maktaba` covering the releases we've made.
- Change codefmt's `.travis.yml` to clone a named Maktaba release.

(I think it's okay to keep depending on Glaive HEAD, since that's not really undergoing much development, and is _only_ needed for the tests.)

malcolmr:
Update the Travis CI build to test against the minimum required version of Maktaba as well as the current HEAD version.

Fixes #35.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: generic
 env:
   matrix:
-    - CI_TARGET=vim
+    - CI_TARGET=vim MAKTABA_VERSION=1.1.0
+    - CI_TARGET=vim MAKTABA_VERSION=master
     - CI_TARGET=neovim
 before_script:
   - sudo apt-get install python3-dev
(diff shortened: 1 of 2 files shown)


malcolmr:
Huh. "Codefmt requires maktaba version 1.1.0.You have maktaba version 1.1.0.
Please update your maktaba install."

I don't think that's working quite as we'd like :-(.

I also see "E118: Too many arguments for function: maktaba#flags#AddCallback", which I suspect means that we actually want something later than 1.1.0.

malcolmr:
Oh, that's because `maktaba#IsAtLeastVersion()` was broken in 1.1.0 (fixed in 1.1.1).

In any case, it looks like we actually now depend upon 1.8.0 due to the `fire_immediately` use added in 92960a1.

dbarnett:
Heh, good thing we're testing now!

LGTM.

malcolmr:
Hmm, it failed with 1.8.0 as well, though it's not exactly clear why. I'll have to see if I can reproduce.

malcolmr:
I pushed a new change that only differed in comments, and tt looks like the Vim failure above was a flake. On the other hand I did spot that I'd completely broken the neovim build; now fixed.

dbarnett:
There's a retry button in the Travis interface if you ever suspect flakes again.

PR #38, opened 2015-03-25
Set the validator for formatters at plugin load, not in autoload.

### LINKED ISSUE #37: Registering external formatters immediately after load breaks codefmt
In `autoload/codefmt`, we have:

```VimL
let s:plugin = maktaba#plugin#Get('codefmt')
let s:plugin_root = expand('<sfile>:p:h:h:h')
let s:registry = s:plugin.GetExtensionRegistry()

call s:registry.SetValidator('codefmt#EnsureFormatter')
```

If the first time we enter this, we already have extensions registered for codefmt, then the call to `SetValidator()` will end up executing `codefmt#EnsureFormatter`.  But that function doesn't yet exist, and codefmt's autoload hasn't yet finished loading, so we'll load it again, and try to loop forever:

```
Vim(call):E132: Function call depth is higher than 'maxfuncdepth'
```

In practice, you can trigger this by installing codefmt, adding an extension via Maktaba, and then running (e.g.) `:FormatCode`, which triggers the autoload.

The simple solution is to move the `SetValidator()` call down to after where `EnsureFormatter()` has been defined (and remove `s:plugin_root`, which is unused).  However, I wonder whether we should instead call `SetValidator()` from `instant/register.vim` or similar?  It seems odd that we'd defer setting up the validator until the first call to a `codefmt#` function.

(This also means that the default formatters won't be registered until that first call, though that's not actually observable, since they're always lower-priority that ones added externally.)

malcolmr:
Fixes #37.


plugin/register.vim
```diff
@@ -0,0 +1,22 @@
...
+" limitations under the License.
+
+let [s:plugin, s:enter] = maktaba#plugin#Enter(expand('<sfile>:p'))
+if !s:enter
+  finish
+endif
+
+
+let s:registry = s:plugin.GetExtensionRegistry()
+call s:registry.SetValidator('codefmt#EnsureFormatter')
```
dbarnett: If this file is called register.vim, it should also register the formatters. That way we'd also avoid weirdness like warnings about the built-in formatters being invalid long after startup.

vroom/main.vroom
```diff
@@ -150,78 +150,3 @@ Similarly, you'll see an error if an explicit formatter name isn't recognized.
 
   :FormatCode nonexistentformatter
   ~ "nonexistentformatter" is not a supported formatter.
-
-Formatters are registered with maktaba's extension registry. Each formatter
```
dbarnett: Add extensions.vroom into the list at the top of this file.

plugin/register.vim
dbarnett: FYI, you can also override the generated help for the flag by defining it explicitly in instant/flags.vim with `@flag` doc comments. "whether plugin/register.vim should be loaded" isn't too bad, but if you wanted to say anything more about it in the docs that would give you a place to do so.

dbarnett: Oh, and the nontrivial code for registering the formatters can still live in autoload. The code here could just be something like:
```vim
call s:Registry.AddExtension(codefmt#GetClangFormatFormatter())
call s:Registry.AddExtension(codefmt#GetGofmtFormatter())
…
```

dbarnett:
LGTM. Only significant question is whether to actually register formatters in register.vim / rename the file.

vroom/main.vroom
malcolmr: Done.

plugin/register.vim
malcolmr: Now registering the formatters.

malcolmr:
PTAL.

autoload/codefmt.vim
```diff
@@ -98,18 +95,20 @@ function! codefmt#EnsureFormatter(formatter) abort
...
-if !exists('s:js_beautify')
-  let s:js_beautify = {
+function! codefmt#GetJsBeautifyFormatter() abort
+  let l:formatter = {
       \ 'name': 'js-beautify',
       \ 'setup_instructions': 'Install js-beautify ' .
           \ '(https://www.npmjs.com/package/js-beautify).'}
 
-  function s:js_beautify.IsAvailable() abort
+  function l:formatter.IsAvailable() abort
```
dbarnett: Wow, does this actually work? I thought trying to nest function definitions in vim did something weird like ending both the inner and outer function at the first `:endfunction`.

If it does work properly I'm fine with it. Otherwise you can just add `#GetJsBeautifyFormatter` as a small wrapper that returns `s:js_beautify`.

vroom/extensions.vroom
```diff
@@ -0,0 +1,75 @@
+Codefmt can be extended to support new formatters, using maktaba's extension
+registry.
```
dbarnett: Could add some more details about what this means, why you would want to extend it vs. send a pull req to codefmt, etc., or link to those details. Don't want to block this bugfix on documentation details, though.

autoload/codefmt.vim
malcolmr: I hadn't seen this syntax at all before, so I just made the change optimistically.  It seems to work; as you say, if we find out that it doesn't work in some cases, we can easily switch the syntax.

dbarnett:
LGTM as long as it works properly wrt the nested function thing. Seems to be passing all tests, at least.

FYI, I just pushed a small change to .travis.yml so the Travis builds won't block for ten minutes waiting for the neovim job to time out.

PR #40, opened 2015-03-30
Remove testing hacks and maktaba version check from bootstrap.vim

### LINKED ISSUE #39: Maktaba minimum version only enforced when using bootstrap.vim
We have some code to require a minimum maktaba version in bootstrap.vim, but if you're not using bootstrap.vim then you'll never get any warning that your maktaba version is too old. We should move that check somewhere it will get executed for all users. instant/flags.vim would probably be a good place.

dbarnett:
Moves testing setup code out of bootstrap.vim (because it's only supposed to get the plugin installed, not try to install glaive from custom paths, etc.). Also moves the maktaba version check so it will trigger for all users and not just any users that source bootstrap.vim.

Fixes #39.

### DIFF
--- a/vroom/setupvroom.vim
+++ b/vroom/setupvroom.vim
@@ -1,0 +1,23 @@
+" Codefmt does not support compatible mode.
+set nocompatible
+
+" Install the codefmt plugin.
+let s:repo = expand('<sfile>:p:h:h')
+execute 'source' s:repo . '/bootstrap.vim'
+
+" Install Glaive from local dir.
+let s:search_dir = fnamemodify(s:repo, ':h')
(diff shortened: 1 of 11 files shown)


malcolmr:
LGTM after the above.

vroom/setupvroom.vim
```diff
@@ -0,0 +1,41 @@
+" Copyright 2015 Google Inc. All rights reserved.
```
dbarnett: Added a copyright notice here. Honestly, I hate having them in every file, but seems like it ought to be here in this file.

malcolmr:
(still lgtm)

Issue #45, opened 2015-07-08
Repeat behavior is unexpected and, well, annoying


luan:
Say you have `:AutoFormatBuffer` enabled and you're editing file. Then you change a line, save it, because why not. You then go to another line and hit `.` to replay that change, it reformats the code.
In my opinion it's pretty clear that the behavior is weird in this situation. I honestly don't see when having code formatting be something that is repeatable via `.` is useful but I'd be interested to hear the reasoning behind it.

https://github.com/google/vim-codefmt/blob/master/autoload/codefmt.vim#L450

I'd send a PR removing that or making it configurable, but I though I'd open a conversation first.
What do y'all think?

dbarnett:
If the format-on-save is replaying with `.`, I'd call that a bug. Are you saying that's what's happening, or are you formatting explicitly before hitting dot?

As far as whether it's useful at all, it's probably a fringe feature. I think it least when formatting via mappings it's theoretically useful, like "format this code block" and then see another block you want formatted and hit `.` to format that as well. I guess text objects are the only case where it really seems useful to me, though. I can't imagine any scenario where I'd expect to format N lines, move the cursor, and then repeat.

fowles:
I agree that AutoFormatBuffer should not repeat, but I think an explicit format command should (and I use that ability)

luan:
I agree with the both of you. and yes `AutoFormatBufffer` currently repeats.

fowles:
If you send the PR, we will be happy to merge it in (modulo signing the google contribution form)

luan:
Sending the PR and closing this for now.

PR #46, opened 2015-07-08
AutoFormatBuffer no longer sets repeat

luan:
Say you have :AutoFormatBuffer enabled and you're editing file. Then you change a line, save it, because why not. You then go to another line and hit . to replay that change, it reformats the code.
In my opinion it's pretty clear that the behavior is weird in this situation. I honestly don't see when having code formatting be something that is repeatable via . is useful but I'd be interested to hear the reasoning behind it. [#45]

This fixes it.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -428,7 +428,7 @@ endfunction
 
 ""
 " Applies [formatter] to the current buffer.
-function! codefmt#FormatBuffer(...) abort
+function! codefmt#FormatBuffer(repeat, ...) abort
   let l:formatter = a:0 >= 1 ? s:GetFormatter(a:1) : s:GetFormatter()
   if l:formatter is# 0
     return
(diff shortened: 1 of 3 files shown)


luan:
I signed it!

autoload/codefmt.vim
```diff
@@ -446,8 +446,10 @@ function! codefmt#FormatBuffer(...) abort
     call maktaba#error#Shout('Error formatting file: %s', v:exception)
   endtry
 
-  let l:cmd = ":FormatCode " . l:formatter.name . "\<CR>"
-  silent! call repeat#set(l:cmd)
+  if a:repeat
+    let l:cmd = ":FormatCode " . l:formatter.name . "\<CR>"
```
dbarnett: You could just put this part in a private function, like:
```vim
""
" @private
function! codefmt#FormatBufferAndSetRepeat(...) abort
  call call('codefmt#FormatBuffer', a:000)
  let l:cmd = ":FormatCode " . l:formatter.name . "\<CR>"
  silent! call repeat#set(l:cmd)
endfunction
```

Otherwise it's a weird signature for `codefmt#FormatBuffer`.

luan:
@dbarnett I agree with your point. Furthermore, looking closer to this I think the functions should not be coupled to the command repetition behavior. I updated the PR extracting that login into command helper functions. Let me know what you think.

dbarnett:
Love it!

PR #49, opened 2015-07-14
Add support for formatting java if >= 3.6

### LINKED ISSUE #43: Add formatter for Java
Codefmt should have support for formatting Java source code if/when a Java formatting tool exists.

It's possible something hacky could be done to organize Java imports without any kind of full-fledged Java formatting tool.

fowles:
Fixes #43


dbarnett:
Seems fine. I'll wait to hear back from @malcolmr before an official LGTM.

Re the concern about precedence of multiple Java formatters, I think like @fowles said codefmt's precedence system should get us the behavior we need.

Issue #27, opened 2015-01-31
Allow configuring tools to inherit vim's indent and line length settings


dbarnett:
Since vim already has settings that correspond to indent size and max line length, it would be great to have an option to hook those up to automatically match vim's 'shiftwidth' and 'textwidth' settings, since I've already configured formatting settings once and shouldn't be forced to repeat myself. There could be a single global setting to enable the feature and a buffer-local override so you could still work around corner cases.

The indent and line length settings will vary from formatter to formatter and support will need to be implemented per-formatter. It could either be explicitly part of the API or formatters could just check the relevant preferences and vim settings themselves in the formatter code.

dbarnett:
My comments from unrelated PR #54:

> My naive preference would be to
> a) Have a global (+local?) option to enable sending shiftwidth, textwidth, and other editor settings to formatters, and
> b) Give custom .clang-format rules higher precedence.
> 
> The reason is that a lot of casual vim users don't configure shiftwidth, textwidth, etc. properly but will still expect codefmt to not cause formatting "regressions", like formatting based on their broken settings instead of custom project configuration.

> Side note: http://editorconfig.org/ looks interesting and I'm wondering if we can consolidate some configuration with that somehow.

My concern is that it adds a new way for users to accidentally get formatting wrong. I'm just hoping to add an explicit opt-in or some heuristics to ensure the shiftwidth we're passing through to formatters is more trustworthy than the formatters' defaults first.

dbarnett:
I gave this idea some more thought and I've decided this shouldn't be automated (though it could still be possible to manually configure flags to pass to formatters as in #66 or #24). You shouldn't get different formatting behavior if you invoke a tool from inside vim vs. standalone. EditorConfig is the ideal solution and Vim shouldn't be in the business of brokering style settings between configs and tools. Formatters tools should be able to detect and respect `.editorconfig` settings (e.g., google/yapf#826 for yapf).

See editorconfig/editorconfig-vim#141 for some additional context.

Closing this bug, but feel free to discuss further if you disagree.

PR #54, opened 2015-08-31
feat: add luaformatter formatter

actionless:


dbarnett:
The shiftwidth change should be a separate PR. Also looks like your vroom tests need to be fixed; they're failing in Travis CI. Besides that, LGTM.

actionless:
do you have any idea why lines are not matching in the test?

dbarnett:
@actionless Looks like it's just typos in the vroom. I see the same when running locally. I think you wanted `:2,3FormatLines` instead of `:2,2FormatLines`. Are you able to run vroom locally?

vroom/luaformatter.vroom
```diff
@@ -0,0 +1,123 @@
...
+      c=1234/2+i
+             end
+  @end
+
+NOTE: the luaformatter formatter does not natively support range formatting,
+so there are certain limitations like not being able to format misaligned
+braces.
+
+
+Shitftwidth will be used for indentation:
```
dbarnett: Unfortunate typo here.

actionless:
i will commit fixes later today

but i also have another question -- how it would be better to handle shiftwidth in clang formatter -- i am thinking about changing default formatting style [1] to `{IndentWidth=`+`&shiftwidth`+`}` but could it be a behavior breakage then if using `.clang-format` file?


[1] https://github.com/google/vim-codefmt/blob/master/instant/flags.vim#L63

dbarnett:
My naive preference would be to
  a) Have a global (+local?) option for sending shiftwidth, textwidth, and other editor settings to formatters, and
  b) Give custom .clang-format rules higher precedence.

The reason is that a lot of casual vim users don't configure shiftwidth, textwidth, etc. properly but will still expect codefmt to not cause formatting "regressions", like formatting based on their broken settings instead of custom project configuration.

OTOH, if you have the editor configured with different settings from your formatters, you're going to have a bad time… It would be nice to just make vim's settings canonical and maybe warn for discrepancies.

Side note: http://editorconfig.org/ looks interesting and I'm wondering if we can consolidate some configuration with that somehow.

malcolmr:
(For a) above, see #27)

actionless:
>Side note: http://editorconfig.org/ looks interesting and I'm wondering if we can consolidate some configuration with that somehow.

i think it should be some separate plugin which will set vim shiftwidth based on config file and codefmt plugin should use those params from vim itself, ie not to depend on any third-party plugin or config file

dbarnett:
I wasn't suggesting an explicit plugin dependency, only that if https://github.com/editorconfig/editorconfig-vim is installed we should cooperate w/ it and avoid any conflicts, and possibly recommend editorconfig on our end to get sane default configuration. It may be as simple as inheriting vim's settings (#27) without needing to know how those were configured, but it's possible we'd also want to treat them as higher precedence if they came from editorconfig since editorconfig users are less likely to have config they don't expect/understand.

actionless:
i will rebase this one after #27 will be implemented then

Kashomon:
I'm going to close this commit since been outstanding for long. It appears to me that the problem was that the the PR tried to do too much: Add a lua formatter (good), change sw (seems reasonable), but they should be in separate PRs.

dbarnett:
> then i will rebase this one after #27 will be implemented

I don't see any reason you would block the luaformatter implementation on the shiftwidth change. Why not just remove that part of the change and follow up in #27, and here fix up the tests and get it submitted?

> but i still can't understand what is the problem to use vim's &shiftwidth -- if someone is using editorconfig plugin then his shitftwidth will be set by it

Commenting in #27 about that. You are correct that shiftwidth would come from editorconfig for users that have it installed and configured. My problem is that for everyone else, shiftwidth adds a new way for users to accidentally get formatting wrong.

PR #56, opened 2015-09-19
Add yapf

shaform:
Add https://github.com/google/yapf


vroom/yapf.vroom
```diff
@@ -0,0 +1,70 @@
...
+  :Glaive codefmt yapf_executable='yapf'
+
+
+You can format any buffer with yapf specifying the formatter explicitly.
+
+  @clear
+  % if True: pass
+
+  :FormatCode yapf
+  ! yapf .* .*
```
fowles: one .* should be sufficient

shaform: Fixed. Thanks!

fowles:
Thanks for the great addition.

PR #59, opened 2015-10-10
Pass -sort-includes to clang-format 3.8 and above.

### LINKED ISSUE #57: Consider using -sort-includes for clang-format
I notice that clang-format has just added a `-sort-includes` option (presumably to what will become clang-format 3.8, but we should verify that); see http://llvm.org/viewvc/llvm-project?view=revision&revision=248367.

This is passed by default by clang-format's own example (Python-based) integration (which we don't use). We should work out whether include sorting (assuming you format that region) is something that makes sense to enable by default. (Yes?)

We probably want to wait until clang-format 3.8 is actually out before adding this, though.

malcolmr:
Fixes #57.

### DIFF
--- a/vroom/clangformat.vroom
+++ b/vroom/clangformat.vroom
@@ -189,11 +189,10 @@ For clang-format versions 3.8 and above, it will sort #include lines (using the
   |#include <stdio.h><CR>
   |int f() { int i = 1; return 1234567890; }
 
-  :goto 1
   :FormatCode clang-format
   ! clang-format-3.7 --version .*
   $ clang-format version 3.7.0 (tags/testing)
-  ! clang-format-3.7 -style file -lines 1:3 -cursor 0 2>.*
+  ! clang-format-3.7 -style file -lines 1:3 -cursor \d+ 2>.*
   $ { "Cursor": 0 }
   $ #include <stdlib.h>
   $ #include <stdio.h>
@@ -214,7 +213,6 @@ For clang-format versions 3.8 and above, it will sort #include lines (using the
   |#include <stdio.h><CR>
   |int f() { int i = 1; return 1234567890; }
 
-  :goto 1
   :FormatCode clang-format
   ! clang-format --version .*
   $ clang-format version 3.8.0 (tags/testing)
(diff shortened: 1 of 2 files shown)


malcolmr:
I was a bit unsure about doing this before clang-format 3.8 was released, but it seems reasonably safe, and allows anyone using trunk versions of clang-format to start using it already.

vroom/clangformat.vroom
```diff
@@ -144,7 +144,7 @@ that clang-format has a version >= 3.4).
   :FormatCode clang-format
   ! clang-format-3.3 --version .*
   $ clang-format version 3.3.0 (tags/testing)
-  ! clang-format-3.3 -style file .*2>.*
+  ! clang-format-3.3 -style file -lines 1:1 2>.*
```
malcolmr: Not strictly needed for this change, but I changed this assertion because I noticed that the previous pattern would have also allowed -cursor to appear here, even though we're trying to test that it doesn't. (And we need to do something similar below.)

vroom/clangformat.vroom
```diff
@@ -180,6 +180,62 @@ that clang-format has a version >= 3.4).
...
+For clang-format versions 3.8 and above, it will sort #include lines (using the
+-sort-includes option).
+
+  :Glaive codefmt clang_format_executable='clang-format-3.7'
+  @clear
+  % #include <stdlib.h><CR>
+  |#include <stdio.h><CR>
+  |int f() { int i = 1; return 1234567890; }
+
+  :goto 1
```
dbarnett: Huh, now I know of at least 4 different ways to go to the beginning of line 1: `1gg`, `1G`, `:1`, and `:goto 1`. Since `:goto 1` is byte-oriented, I might avoid it except when using it with other byte counts; line-oriented operations are more versatile in common practice.

dbarnett:
LGTM.

vroom/clangformat.vroom
malcolmr: Good point.

In fact, in this test, we don't care about the cursor position at all, so I'll remove them (and replace `-cursor 0` with `-cursor \d+` in the expectation).

PR #62, opened 2015-11-02
Fixes #61: Catch yapf errors instead of overwriting file

### LINKED ISSUE #61: yapf code formatting deletes file if it does not parse
If you have a file that does not parse correctly by Python, yapf returns an error, and does not print anything to stdout. codefmt reads the stdout from yapf, and replaces the vim buffer with it, which means that a file with the below contents (note the missing end quote)

```
print 'hello
```

will be emptied when codefmt is run.

achalddave:
This isn't perfect, since it doesn't try to parse the yapf output, but it at least won't overwrite the file if there's an error.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -437,15 +437,14 @@ function! codefmt#GetYAPFFormatter() abort
     let l:cmd = [l:executable, '--lines=' . a:startline . '-' . a:endline]
     let l:input = join(l:lines, "\n")
 
-    try
-      let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
-      let l:formatted = split(l:result.stdout, "\n")
+    let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call(0)
+    if v:shell_error == 1 " Indicates an error with parsing
+      call maktaba#error#Shout('Error formatting file: %s', l:result.stderr)
+      return
+    endif
+    let l:formatted = split(l:result.stdout, "\n")
 
-      call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
-    catch /ERROR(ShellError):/
-      call maktaba#error#Shout('Error formatting file: %s', v:exception)
-    endtry
-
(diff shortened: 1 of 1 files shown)


achalddave:
Just signed the CLA.

autoload/codefmt.vim
```diff
@@ -437,12 +437,17 @@ function! codefmt#GetYAPFFormatter() abort
     let l:cmd = [l:executable, '--lines=' . a:startline . '-' . a:endline]
     let l:input = join(l:lines, "\n")
 
-    let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call(0)
-    let l:formatted = split(l:result.stdout, "\n")
+    try
+      let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
+      let l:formatted = split(l:result.stdout, "\n")
+
+      let l:full_formatted = l:formatted
```
dbarnett: Could you drop this reassignment while you're in here? AFAICT it serves no purpose.

achalddave: Oops, didn't notice that. Done.

achalddave:
This is actually wrong, and I don't think I'm familiar enough with vimscript to fix it. Specifically, yapf returns error code 2 if the error is changed, which means I have to catch the ShellError exception, then check if the error code is 2, and do the same formatting logic as in the try. Something like

```
    try
      let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
    catch /ERROR(ShellError):/
      if v:shell_error == -1
        call maktaba#error#Shout('Error formatting file: %s', v:exception)
      endif
    endtry

    call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
```

doesn't work since `l:result` is scoped to within the try block... I don't know how to declare result to have scope outside the try block in vimscript.

achalddave:
Does this look good? I've been using it for the past few days and it seems to work fine.

dbarnett:
Yep, looks good. Thanks!

PR #63, opened 2015-11-17
Fix clang version check

achalddave:


malcolmr:
Can we add (or change) a test in `vroom/clangformat.vroom` so that we can see what this is fixing?

achalddave:
Oops. This was my bad due to a few odd steps that simultaneously led to me submitting this PR

- The version of clang-format installed by brew reports itself as 3.8, but does not having -sort-includes. I had this version installed, but -sort-includes was broken, so
- I reinstalled clang-format separately which was reported as version 3.7. However, a background vim process I had open must have still had the earlier binary loaded, and was still passing -sort-includes. I assumed this was a bug in the vim-codefmt code to check the version.
- I saw a <= instead of < in the short circuit above (which would pass a 3.7 version through a 3.8 check); and of course, fixing it and restarting vim made things work (since vim now reloaded the v3.7 binary). Thus, I assumed it was a bug in codefmt and hastily sent this PR without looking it over again...

tl;dr - My bad, this PR is buggy.

Issue #67, opened 2016-01-26
Add dartfmt formatter for Dart language


dbarnett:
Dart has a formatter: https://github.com/dart-lang/dart_style. Codefmt should support invoking dartfmt to format Dart code.

DavidSouther:
I'll take a stab at that.

Kashomon:
This was added.

PR #68, opened 2016-01-26
Added dartfmt code formatter.

DavidSouther:
This implements #67


DavidSouther:
"It works for me!"

I don't know how to run vroom, so let's hope Travis tells me I got it right the first time.

DavidSouther:
This implements #67

autoload/codefmt.vim
```diff
@@ -326,6 +327,42 @@ function! codefmt#GetGofmtFormatter() abort
...
+  endfunction
+
+  function l:formatter.AppliesToBuffer() abort
+    return &filetype is# 'dart'
+  endfunction
+
+  ""
+  " Reformat the current buffer with dartfmt or the binary named in
+  " @flag{dartfmt_executable}. At this time, does not handle ranges of lines.
+  function l:formatter.FormatRange(startline, endline) abort
```
fowles: since you don't support ranges you should implement Format() only

DavidSouther: Done.

fowles:
LGTM, but lets wait on travis

DavidSouther:
Yeah this is totally not going to work. BRB doing it differently.

DavidSouther:
Ok, this actually formats the buffer, rather than the file the buffer was loaded from.

DavidSouther:
Ah - turns out they do, and I was calling it wrong.

DavidSouther:
Boom - works for ranges, on stdin, and puts errors in quickfix!

autoload/codefmt.vim
```diff
@@ -326,6 +326,74 @@ function! codefmt#GetGofmtFormatter() abort
...
+          call add(l:errors, {
+              \ 'filename': @%,
+              \ 'lnum': l:tokens[1] + a:startline - 1,
+              \ 'col': l:tokens[2],
+              \ 'text': l:tokens[3]})
+        endif
+      endfor
+
+      if empty(l:errors)
+        " Couldn't parse gofmt error format; display it all.
```
fowles: s/gofmt/dartformat/

vroom/dartfmt.vroom
```diff
@@ -0,0 +1,68 @@
...
+  @end
+
+The dart filetype will use the dartfmt formatter by default.
+
+  @clear
+  % main(){}
+
+  :set filetype=dart
+  :FormatCode
+  ! dartfmt .*
```
fowles: add a test for range formatting

DavidSouther: dartfmt *really* doesn't like formatting anything less than a complete function or class. Are we sure we want to support range formatting when it will error basically every time someone doesn't do the full file?

autoload/codefmt.vim
```diff
@@ -326,6 +326,74 @@ function! codefmt#GetGofmtFormatter() abort
...
+    return executable(s:plugin.Flag('dartfmt_executable'))
+  endfunction
+
+  function l:formatter.AppliesToBuffer() abort
+    return &filetype is# 'dart'
+  endfunction
+
+  ""
+  " Reformat the current buffer with dartfmt or the binary named in
+  " @flag{dartfmt_executable}, only targetting the range from {startline} to
```
dbarnett: `@flag(dartfmt_executable)` here (parens instead of curly braces).

vroom/dartfmt.vroom
dbarnett: I always suggest at least filing a feature request in the external project to implement context-aware range formatting. Up to you whether you want to support range formatting like this. It gives the option to people who know what they're doing, but could confuse others who don't.

DavidSouther: https://github.com/dart-lang/dart_style/issues/92

autoload/codefmt.vim
DavidSouther: Done.

fowles:
I think we may have lost track of this one (sorry!).  What is the current status of it?

DavidSouther:
Should be fine? Looks like there's no conflicts, tests are passing. There was the open issue of whether dart-lang would handle ranges, but that's still open. dart-lang/dart_style#92

fowles:
@dbarnett I am fine with this, do you have any thoughts?

DavidSouther:
I think I've done this??

I install [google/vimdoc](https://github.com/google/vimdoc), followed its install instructions, and ran it in the directory. Let me know what else I'd need to do to get it done correctly.

dbarnett:
Looks like some unrelated "gn" changes appeared in 742f9de757, and some of the dartfmt changes disappeared. The docs did get updated so it looks like you successfully ran vimdoc, but I don't know where all those other changes came from.

DavidSouther:
Ok, rebased to master, this looks sane.

dbarnett:
(I'm curious… does approving changes grant you permission to merge them yourself with a "Merge pull request" button, or is it just informational?)

PR #71, opened 2016-04-19
Add builtin gn formatter.

sfiera:
Chrome uses BUILD.gn files to build on most platforms. The command
includes a built-in formatter.


sfiera:
No CLA necessary; see http://who/sfiera (my @chromium.org account wasn't yet associated with this account).

PR #77, opened 2017-01-23
Split out all the formatters into separate files for readability.

Kashomon:
Following my previous changes to the README and the buildifier plugin, split out the formatters into separate files. This makes it easier to copy-pasta to create new formatters.


autoload/codefmt/autopep8.vim
```diff
@@ -0,0 +1,97 @@
...
+"
+" Unless required by applicable law or agreed to in writing, software
+" distributed under the License is distributed on an "AS IS" BASIS,
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+" See the License for the specific language governing permissions and
+" limitations under the License.
+
+
+let s:plugin = maktaba#plugin#Get('codefmt')
+let s:registry = s:plugin.GetExtensionRegistry()
```
dbarnett: None of these new files actually use s:registry AFAICT.

Kashomon: Mind if I follow-up?

PR #81, opened 2017-03-02
Add a test for error reporting for buildifier.

malcolmr:


vroom/buildifier.vroom
```diff
@@ -52,3 +52,22 @@ The bzl filetype will use the buildifier formatter by default.
   $ )
 
   :set filetype=
+
+Errors are reported using the quickfix list.
+
+  @clear
+  % #<CR>
+  |java_package(])
```
dbarnett: Nit: I think it's clearer as
```
  % #<CR>
  % java_package(])
```
like https://github.com/google/vroom/blob/master/examples/buffer.vroom does. I had to think more about the vroom syntax in the line continuation form.

malcolmr: Done (I was confused a bit as well), though note that the prevailing style in some of the other test files is for the continuation versions.

PR #96, opened 2018-05-21
fix(js-beautify): Use correct type for javascript filetype.

### LINKED ISSUE #95: jsbeautify does not take 'javascript' as a type
FormatCode fails on JavaScript files.
I am using version 1.7.5 of js-beautify. It takes 'js' instead of 'javascript' for type.

bjagg:
If the filetype is 'Javascript', use 'js' for type parameter to js-beautify.

Fixes #95.

### DIFF
--- a/autoload/codefmt/jsbeautify.vim
+++ b/autoload/codefmt/jsbeautify.vim
@@ -41,7 +41,9 @@ function! codefmt#jsbeautify#GetFormatter() abort
   " @throws ShellError
   function l:formatter.FormatRange(startline, endline) abort
     let l:cmd = [s:plugin.Flag('js_beautify_executable'), '-f', '-']
-    if &filetype != ""
+    if &filetype is# 'javascript'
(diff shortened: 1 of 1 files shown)


autoload/codefmt/jsbeautify.vim
```diff
@@ -41,7 +41,9 @@ function! codefmt#jsbeautify#GetFormatter() abort
   " @throws ShellError
   function l:formatter.FormatRange(startline, endline) abort
     let l:cmd = [s:plugin.Flag('js_beautify_executable'), '-f', '-']
-    if &filetype != ""
+    if &filetype is# 'javascript'
+      let l:cmd = l:cmd + ['--type', 'js']
```
malcolmr: We should also use `--type=js` for the json filetype.

bjagg:
Updated to handle 'json' filetype as well. Thanks!

PR #97, opened 2018-05-21
Improve jsbeautify tests

malcolmr:
Prior to #96 (3dad76a and 8e54240), the js-beautify binary was invoked with the wrong --type argument.

This change verifies the js-beautify arguments for each of the supported filetypes (so that we can verify the fixes above), and also explains what's wrong with the javascript filetype support (it's picked up by clang-format instead).


dbarnett:
LGTM!

PR #99, opened 2018-06-06
Enable clang-format for objc/objcpp files.

### LINKED ISSUE #15: Associate clang-format formatter with C and Obj-C
In clang-format's help, it calls itself "a tool to format C/C++/Obj-C code", but the current AppliesToBuffer check on the clang-format formatter only checks for cpp, proto, and javascript filetypes.

We should enable it as a default formatter for filetypes c, objc, and objcpp so that if people use `:FormatCode` on these filetypes with no explicit formatter, they will get code formatted with clang-format instead of seeing `Not available. codefmt doesn't have a default formatter for this buffer`.

rickyz:
Fixes #15


rickyz:
Fixes #99

PR #104, opened 2018-08-14
Fix buildifier error reporting.

### LINKED ISSUE #103: buildifier integration no longer reports errors correctly
The changes for #100 broke the way that buildifier reports errors, because we assume that the filename is called "stdin", and after #101, it's the value of `-path`:

https://github.com/google/vim-codefmt/blob/4b6cb67/autoload/codefmt/buildifier.vim#L57

This results in e.g.

```
Error formatting file: ERROR(ShellError): Error running: buildifier -path /home/malcolmr/foo/bar/BUILD
/home/malcolmr/foo/bar/BUILD:112:13: syntax error near cc_library
```

malcolmr:
This was broken by 4b6cb67, which caused error messages to contain the filename rather than 'stdin'.

Fixes #103

### DIFF
--- a/autoload/codefmt/buildifier.vim
+++ b/autoload/codefmt/buildifier.vim
@@ -54,7 +54,13 @@ function! codefmt#buildifier#GetFormatter() abort
       " Parse all the errors and stick them in the quickfix list.
       let l:errors = []
       for line in split(v:exception, "\n")
-        let l:tokens = matchlist(line, '\C\v^stdin:(\d+):(\d+):\s*(.*)')
+        if empty(l:fname)
+          let l:fname_pattern = 'stdin'
+        else
+          let l:fname_pattern = escape(l:fname, '\')
+        endif
+        let l:tokens = matchlist(line,
+            \ '\C\v^\V'. l:fname_pattern . '\v:(\d+):(\d+):\s*(.*)')
         if !empty(l:tokens)
           call add(l:errors, {
               \ "filename": @%,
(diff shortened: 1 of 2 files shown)


autoload/codefmt/buildifier.vim
```diff
@@ -54,7 +54,13 @@ function! codefmt#buildifier#GetFormatter() abort
       " Parse all the errors and stick them in the quickfix list.
       let l:errors = []
       for line in split(v:exception, "\n")
-        let l:tokens = matchlist(line, '\C\v^stdin:(\d+):(\d+):\s*(.*)')
+        if empty(l:fname)
+          let l:fname_pattern = 'stdin'
+        else
+          let l:fname_pattern = escape(l:fname, '\')
```
malcolmr: Alternatively we could just make the pattern start out something like `'\C\v[^:]+:\(\d+)...`, and assume that filenames never contained a `:`, but I suspect that wouldn't work everywhere.

dbarnett: I could imagine corner cases with the filename being escaped/expanded/canonicalized that with this approach would lead to the error not being matched at all. It might be worth figuring out how to genericize it properly eventually (not a blocker since this does fix the major issue).

malcolmr: I'm pretty sure that buildifier returns exactly the string we pass to it via `-path` (after all, it doesn't actually get to read the file content: that's piped to stdin). I'll commit this as-is, and we can always improve it if we need to later.

PR #110, opened 2018-12-25
prettier: add other supported filetypes (html, css, etc).

infokiller:


dbarnett:
There are some test failures in https://travis-ci.org/google/vim-codefmt/builds/472093600 that indicate there's an issue with which formatter is used by default for html filetype vs. the test expectations. Right now prettier is registered to take precedence over jsbeautify based on registration order (see https://github.com/google/vim-codefmt/blob/2147649088e977852c168a2173f13e0e04d555ed/plugin/register.vim#L24) but jsbeautify.vroom line 78 is expecting the js-beautify formatter to have precedence for html. Which precedence makes sense?

You'll need to either update the precedence or update jsbeautify.vroom if there's a rationale for precedence being that way.

infokiller:
Sorry, didn't notice that failure.
I think it makes more sense to have jsbeautify take precedence so that the behavior stays the same when users update the plugin.
I changed it and now it should pass.

PR #121, opened 2019-07-06
Use vim's builtin format mapping

tbodt:


dbarnett (commented):
I like providing a simple hook for people who want to override formatexpr and an example of how. It would be pretty aggressive behavior to unconditionally override formatexpr, though. Could be based on a setting, but for things like this it's usually more flexible and less brittle to provide the pieces and let the user hook it up themselves. For instance, if they wanted to override formatexpr for some but not all filetypes, they'd come asking how to adapt this to do what they wanted and we wouldn't have a simple clean way to offer them.

What do you think about changing it so the autocmd is in your vimrc and just calls a simple codefmt helper?

dbarnett (changes requested):
I think it needs to be opt-in. See my previous longer comment in Conversation.

tbodt:
Removed autocmd from mappings file, PTAL

autoload/codefmt.vim
```diff
@@ -252,6 +262,23 @@ function! codefmt#GetSupportedFormatters(ArgLead, CmdLine, CursorPos) abort
...
+""
+" Returns whether there is a default formatter available for the current
+" buffer.
+function! codefmt#AvailableInCurrrentBuffer() abort
+  let l:formatters = s:registry.GetExtensions()
+  if !empty(get(b:, 'codefmt_formatter'))
+    let l:Predicate = {f -> f.name ==# b:codefmt_formatter}
+  else
+    let l:Predicate = {f -> f.AppliesToBuffer() && s:IsAvailable(f)}
+  endif
```
dbarnett: For reference: We haven't been using lambdas in any of our plugins yet, but I checked and they were originally added in 7.4.2044 from 2016. That's probably long enough ago. At some point I should probably declare a min supported vim version for the plugin and get it testing against that version, though.

autoload/codefmt.vim
```diff
@@ -252,6 +262,23 @@ function! codefmt#GetSupportedFormatters(ArgLead, CmdLine, CursorPos) abort
   return join(l:groups[0] + l:groups[1] + l:groups[2], "\n")
 endfunction
 
+""
+" Returns whether there is a default formatter available for the current
```
dbarnett: Could you make this `@public` and regenerate the vimdoc?

It should have a sample of how to use it somewhere in the help docs, but for that part I'm happy to merge it w/o and add the docs after.

dbarnett (changes requested):
LG to merge besides getting the functions into the generated help docs. LMK if you need any help with that part.

autoload/codefmt.vim
tbodt: Done (two years later :)

dbarnett:
Thanks!

> Removed autocmd from mappings file, PTAL

Looks like the autocmd is still in the mappings file? I'd suggest to instead restructure a bit to define a `codefmt#SetFormatExprIfAvailable` and then replace the unconditional autocmd with some instructions in the help file / README to recommend putting these lines in your vimrc:
```vim
augroup codefmt_formatexpr
  autocmd!
  autocmd BufEnter * call codefmt#SetFormatExprIfAvailable()
augroup END
```

Issue #123, opened 2019-08-06
Unable to run clang-format in docker


funous:
I am running vim on my host machine, while all my C++ projects are in their corresponding docker containers. I'd like to be able to configure vim-codefmt to run in the corresponding docker container, however it is not possible at the moment. `clang_format_executable` can only be set to an executable name with no arguments, however I'd like to be able to set it to a command that does include arguments (such as `docker exec -i <my-container> clang-format`, similar to what can be done with `google_java_executable`).

funous:
Resolved by #124

PR #124, opened 2019-08-06
Add support for an arbitrary clang_format command with arguments

funous:
Solves #123 by applying handling of `clang_format_executable` similar to the existing handling of `google_java_executable`

### DIFF
--- a/autoload/codefmt/clangformat.vim
+++ b/autoload/codefmt/clangformat.vim
@@ -18,12 +18,13 @@ let s:plugin = maktaba#plugin#Get('codefmt')
 
 function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
   if !exists('s:clang_format_version')
-    let l:executable = s:plugin.Flag('clang_format_executable')
-    if codefmt#ShouldPerformIsAvailableChecks() && !executable(l:executable)
+    let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray(
+          \ 'clang_format_executable')
+    if codefmt#ShouldPerformIsAvailableChecks() && !executable(l:cmd[0])
       return 0
     endif
 
-    let l:syscall = maktaba#syscall#Create([l:executable, '--version'])
+    let l:syscall = maktaba#syscall#Create(l:cmd + ['--version'])
     " Call with throw_errors disabled because some versions of clang-format
     " misbehave and return exit code 1 along with the successful version
     " output (see https://github.com/google/vim-codefmt/issues/84).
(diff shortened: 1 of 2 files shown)


autoload/codefmt/clangformat.vim
```diff
@@ -91,9 +108,9 @@ function! codefmt#clangformat#GetFormatter() abort
       return
     endif
 
-    let l:cmd = [
-        \ s:plugin.Flag('clang_format_executable'),
-        \ '-style', l:style]
+    " Split the command on spaces, except when there's a proceeding \
```
malcolmr: I know this is copy/pasted, but I think you want "preceding" here.

autoload/codefmt/clangformat.vim
```diff
@@ -19,8 +19,16 @@ let s:plugin = maktaba#plugin#Get('codefmt')
...
     let l:executable = s:plugin.Flag('clang_format_executable')
-    if codefmt#ShouldPerformIsAvailableChecks() && !executable(l:executable)
-      return 0
+    if codefmt#ShouldPerformIsAvailableChecks()
+      if !executable(l:executable)
+        return 0
+      elseif !empty(l:executable) && l:executable isnot# 'clang-format'
+        " The user has specified a custom formatter command. Hope it works.
+        " /shrug.
+        return 0
```
malcolmr: This (and the branch immediately below) aren't correct: in both cases (custom executable and 'clang-format' being executable), we want to continue to call --version to check the reported version number.

The only case we want to short-circuit is the non-custom-executable case where the executable can't be run.

funous: I've updated the logic

autoload/codefmt/clangformat.vim
funous: Fixed

malcolmr:
I've fixed the problem with Travis in #125 (sorry about that); could you rebase (and so trigger the CI test again) so that we can check that the tests still pass?

agriffis:
This is similar to what I needed in #136 to run `npx prettier` instead of `prettier`. The approach I took there was to support `prettier_executable` as a list in addition to a simple string, using a helper function for the boilerplate code.

We might want to consider the same approach here, so that we don't end up with bug reports later on because of splitting words without respect to shell quoting.

Assuming #136 goes through, I'll loop back to this PR with a diff for discussion.

funous:
I've updates this to use the new `codefmt#formatterhelpers#ResolveFlagToArray` to handle the arguments, introduced in https://github.com/google/vim-codefmt/pull/138. This makes the implementation cleaner.

PR #125, opened 2019-08-12
Start xvfb using 'services' rather than a run script.

malcolmr:
This is necessary now that Travis has switched from Trusty to Xenial by
default.

(See https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-services)


dbarnett:
Ah nice, thanks!

Presumably we need to set this up in the canonical example at https://github.com/google/vroom/blob/master/.travis.yml and the Travis configs for all the other plugins as well?

malcolmr:
Thanks. Yes, I'll do the same for the other repositories now.

PR #130, opened 2019-11-11
feat: add zprint formatter for clojure

agriffis:
This pull request adds a formatter for Clojure using zprint.

zprint unfortunately lacks the ability to format a range, so we implement `:FormatLines` by sending only the lines to be formatted. I've opened a PR to add range formatting to zprint, see https://github.com/kkinnear/zprint/issues/122

Doc is updated by vimdoc, with a special section `*codefmt-mappings-zprint*` suggesting an alternative mapping for top-level forms which depends on also having [vim-sexp](https://github.com/guns/vim-sexp) installed:

    autocmd FileType clojure nmap <buffer> <silent> <leader>== <leader>=iF


.travis.yml
```diff
@@ -20,6 +22,9 @@ before_script:
   - sudo dpkg -i ./vroom_0.12.0-1_all.deb
   - git clone -b ${MAKTABA_VERSION} https://github.com/google/vim-maktaba.git ../maktaba/
   - git clone https://github.com/google/vim-glaive.git ../glaive/
+  - wget https://github.com/kkinnear/zprint/releases/download/${ZPRINT_VERSION}/zprintl-${ZPRINT_VERSION}
+  - sudo mv zprintl-${ZPRINT_VERSION} /usr/local/bin/zprint
+  - sudo chmod +x /usr/local/bin/zprint
```
dbarnett: I think the GNU `install` command is supposed to move and chmod files in one go. Can you check if that will work here?

autoload/codefmt/zprint.vim
```diff
@@ -0,0 +1,82 @@
...
+" Unless required by applicable law or agreed to in writing, software
+" distributed under the License is distributed on an "AS IS" BASIS,
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+" See the License for the specific language governing permissions and
+" limitations under the License.
+
+
+let s:plugin = maktaba#plugin#Get('codefmt')
+
+let s:supported_filetypes = ['clojure']
```
dbarnett: Style nit: Is this meant to be constant or updateable? If constant, consider capitalizing and doing `lockvar!` since vimscript functions make it easy to accidentally modify lists otherwise (and you'll probably want an `if exists()` guard since this is an autoload file). Otherwise, you could attach it to `s:plugin.globals`.

autoload/codefmt/zprint.vim
```diff
@@ -0,0 +1,82 @@
...
+            \ string(l:ZprintOptions))
+    endif
+    let l:cmd = [s:plugin.Flag('zprint_executable')]
+    call extend(l:cmd, l:zprint_options)
+
+    call maktaba#ensure#IsNumber(a:startline)
+    call maktaba#ensure#IsNumber(a:endline)
+    let l:lines = getline(1, line('$'))
+
+    " zprint doesn't support formatting a range of lines, so format the range
```
dbarnett: Can you follow up with zprint to request real range formatting and post a pointer to an issue in their tracker here? For a functional language, hacked range formatting is going to be especially brittle.

vroom/zprint.vroom
```diff
@@ -0,0 +1,95 @@
...
+  $   (cond nil 1
+  $         :else 2))
+  (defn x
+    []
+    (cond nil 1
+          :else 2))
+  (defn x [] (cond nil 1 :else 2))
+  @end
+
+It is also the default formatter for the clojure file type.
```
dbarnett: The descriptions would read better if you reordered the sections so this `It is also the default formatter for` section is right after the `You can format any buffer` section.

instant/flags.vim
```diff
@@ -115,3 +115,18 @@ call s:plugin.Flag('prettier_options', [
 ""
 " The path to the prettier executable.
 call s:plugin.Flag('prettier_executable', 'prettier')
+
+""
+" Whether and where to change directory before calling zprint. If this is true,
```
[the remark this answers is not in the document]
agriffis: Yeah, I got rid of this option in a later commit, with the same thinking. I was planning to squash the commits when I remove the draft status on the PR

vroom/zprint.vroom
```diff
@@ -0,0 +1,81 @@
...
+  :call maktaba#test#Override('repeat#set', 'FakeRepeat')
+
+  :call codefmt#SetWhetherToPerformIsAvailableChecksForTesting(0)
+
+
+The zprint formatter expects the zprint executable to be installed on your system.
+
+  % (defn x [] (cond nil 1 :else 2))
+  :FormatCode zprint
+  ! zprint .*
```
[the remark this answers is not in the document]
agriffis: Thanks, I eventually figured this out too. Tests are passing now.

autoload/codefmt/zprint.vim
agriffis: That's from `autoload/codefmt/prettier.vim` but it would make more sense to use the simpler `AppliesToBuffer` function from `autoload/codefmt/rustfmt.vim` since there's just one filetype.

autoload/codefmt/zprint.vim
agriffis: Will do

vroom/zprint.vroom
agriffis: Will do

.travis.yml
agriffis: I had considered that, but GNU `install` copies rather than moves and that seemed messy. I don't have strong feelings about it—just let me know what you'd like.

agriffis: Done

autoload/codefmt/zprint.vim
agriffis: Done

autoload/codefmt/zprint.vim
agriffis: Done, but I'll also update the PR description to include this link before removing the draft status. https://github.com/kkinnear/zprint/issues/122

vroom/zprint.vroom
agriffis: I didn't reorder exactly like that, but I reworked the narrative so it's clearer and doesn't have a continuity dependency between the sections. (I wanted to keep `:FormatCode` and `:FormatLines` next to each other, but previously the test for `:FormatCode` was worded to emphasize calling zprint explicitly.)

agriffis:
@dbarnett Thanks for the review. I've squashed the commits and this should be ready to merge unless you have any other concerns.

agriffis:
If you pull #131 before this, then I'll rebase this PR to resolve the minor merge conflict

agriffis:
Rebased to resolve merge conflict with #131

.travis.yml
```diff
@@ -23,6 +24,8 @@ before_script:
   - sudo dpkg -i ./vroom_${VROOM_VERSION}-1_all.deb
   - git clone -b ${MAKTABA_VERSION} https://github.com/google/vim-maktaba.git ../maktaba/
   - git clone https://github.com/google/vim-glaive.git ../glaive/
+  - wget https://github.com/kkinnear/zprint/releases/download/${ZPRINT_VERSION}/zprintl-${ZPRINT_VERSION}
```
dbarnett: Wait, do you actually need zprint installed for these if you're faking the I/O?

agriffis: I totally did not understand that aspect of vroom. I think the tests will need another pass to straighten this out, but then I think you're right.

agriffis: @dbarnett Rebased now on master with updated tests to fake I/O properly, and removed zprint install from travis.

agriffis:
@dbarnett In case this was lost in code review comments, I updated the tests to use fake I/O fully, and removed the zprint install from travis. Let me know if there's anything else...

autoload/codefmt/zprint.vim
```diff
@@ -0,0 +1,91 @@
...
+"
+" Unless required by applicable law or agreed to in writing, software
+" distributed under the License is distributed on an "AS IS" BASIS,
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+" See the License for the specific language governing permissions and
+" limitations under the License.
+
+""
+" @section Recommended zprint mappings, mappings-zprint
+" @parentsection mappings
```
dbarnett: Note to self: Build a new vimdoc release. I just found that the `@parentsection` directive isn't supported yet in the latest vimdoc 0.6.0 release.

PR #131, opened 2019-12-30
fix: neovim vroom tests

agriffis:
Four problems that were preventing the vroom tests from running on
neovim:

- The vroom version was too old, see
  https://github.com/google/vroom/pull/104

- The maktaba version was too old, see
  https://github.com/google/vim-maktaba/issues/172. I considered
  constraining this to neovim but it applies to recent vim as well (the
  version of vim in travis/xenial is quite old) and the extra version
  checking seems pointless.

- The get-pip.py invocation was outdated.

- For some reason neovim pauses more often with "Hit ENTER" messages.
  Perhaps newer vim would too. In any case, increasing the cmdheight
  solves the problem without needing to sprinkle :silent everywhere.

### DIFF
--- a/instant/flags.vim
+++ b/instant/flags.vim
@@ -33,6 +33,9 @@ endif
 
 
 " Shout if maktaba is too old. Done here to ensure it's always triggered.
+" We need at least 1.12.0 so that maktaba#ensure#IsCallable works on Neovim and
+" recent Vim (newer than is actually in Travis/Xenial).
+" See https://github.com/google/vim-maktaba/issues/173
 if !maktaba#IsAtLeastVersion('1.12.0')
   call maktaba#error#Shout('Codefmt requires maktaba version 1.12.0.')
   call maktaba#error#Shout('You have maktaba version %s.', maktaba#VERSION)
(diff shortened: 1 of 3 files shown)


instant/flags.vim
```diff
@@ -33,8 +33,8 @@ endif
 
 
 " Shout if maktaba is too old. Done here to ensure it's always triggered.
-if !maktaba#IsAtLeastVersion('1.10.0')
-  call maktaba#error#Shout('Codefmt requires maktaba version 1.10.0.')
+if !maktaba#IsAtLeastVersion('1.12.0')
```
dbarnett: Could you also comment what feature needs 1.12?

agriffis: That's the second bullet in the commit message. Maktaba prior to 1.12.0 doesn't work on any relatively modern vim or neovim because of the linked issue. The only reason 1.10.0 works in travis is that [vim on xenial is 7.4.1689](https://packages.ubuntu.com/search?keywords=vim-gnome) which is prior to the breakage.

I guess the other way to approach this is to say that vim-codefmt only requires 1.10.0 in `instant/flags.vim` but the neovim tests require 1.12.0 in `.travis.yml`. Would that be preferable to you?

dbarnett: Yep, the reason is fine, I just meant to have it as a comment in the file itself w/ some kind of self-contained link to the reason.

agriffis: Oh, gotcha

Issue #60, opened 2015-10-23
Consider providing a standard way for formatters to report errors


malcolmr:
Currently, if formatters want to report errors, they update the quickfix list directly. (As shipped, only the gofmt formatter does this, but it would also make sense for e.g. google-java-format, were we to add support for that.) It would be nice to make it easier for formatters to report errors.

Additionally, we have some questions around when formatters should clear the quickfix list (e.g. #58), and I'm not sure how stacked formatters (#44) would work if formatters were managing the quickfix list individually.

I've been wondering whether formatters could pass a structured error list to codefmt somehow, and then codefmt itself could be responsible for merging/populating/hiding the quickfix list as appropriate, consistently.

As a strawman, perhaps `FormatRange()` (etc) could return a list of (line, column, message)? (Or accept a callback instead, if that sounds like a restrictive API? Would the line numbers be relative to the start of the range, for range formatting?)

And/or, since most formatters can get errors by looking for "file:line:column: error" on stderr, it might even be work providing a helper that handled exactly that pattern directly.

dbarnett:
+1

Related: I'm considering an alternative API for formatters (#14) where codefmt passes a handle in and the formatter can perform operations on it more flexibly. It's possibly overkill but I'll bring it up since we're talking about extending the API.

PR #135, opened 2020-01-08
Implement common helper functions for common formatting operations

dbarnett:
Creates helper functions for the boilerplate of common operations in formatter implementations: `codefmt#formatterhelpers#Format({cmd})` and `codefmt#formatterhelpers#AttemptFakeRangeFormatting({startline}, {endline}, {cmd})`. These make conventions more obvious and help any intentional differences from the most common boilerplate version stand out more obviously.

### DIFF
--- a/autoload/codefmt/autopep8.vim
+++ b/autoload/codefmt/autopep8.vim
@@ -67,29 +67,18 @@ function! codefmt#autopep8#GetFormatter() abort
 
     call maktaba#ensure#IsNumber(a:startline)
     call maktaba#ensure#IsNumber(a:endline)
-    let l:lines = getline(1, line('$'))
 
     if s:autopep8_supports_range
-      let l:cmd = [l:executable, '--range', ''.a:startline, ''.a:endline, '-']
-      let l:input = join(l:lines, "\n")
+      call codefmt#formatterhelpers#Format(maktaba#syscall#Create([
+          \ l:executable,
+          \ '--range', string(a:startline), string(a:endline),
+          \ '-']))
     else
-      let l:cmd = [l:executable, '-']
-      " Hack range formatting by formatting range individually, ignoring context.
-      let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
+      call codefmt#formatterhelpers#AttemptFakeRangeFormatting(
+          \ a:startline,
+          \ a:endline,
+          \ maktaba#syscall#Create([l:executable, '-']))
     endif
-
-    let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
-    let l:formatted = split(l:result.stdout, "\n")
-
-    if s:autopep8_supports_range
-      let l:full_formatted = l:formatted
-    else
-      " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
-      let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
-      let l:full_formatted = l:before + l:formatted + l:lines[a:endline :]
-    endif
-
-    call maktaba#buffer#Overwrite(1, line('$'), l:full_formatted)
   endfunction
 
   return l:formatter
--- a/autoload/codefmt/autopep8.vim
+++ b/autoload/codefmt/autopep8.vim
@@ -69,15 +69,15 @@ function! codefmt#autopep8#GetFormatter() abort
     call maktaba#ensure#IsNumber(a:endline)
 
     if s:autopep8_supports_range
-      call codefmt#formatterhelpers#Format(maktaba#syscall#Create([
+      call codefmt#formatterhelpers#Format([
           \ l:executable,
           \ '--range', string(a:startline), string(a:endline),
-          \ '-']))
+          \ '-'])
     else
       call codefmt#formatterhelpers#AttemptFakeRangeFormatting(
           \ a:startline,
           \ a:endline,
-          \ maktaba#syscall#Create([l:executable, '-']))
+          \ [l:executable, '-'])
     endif
   endfunction
 
--- a/autoload/codefmt/buildifier.vim
+++ b/autoload/codefmt/buildifier.vim
@@ -49,7 +49,7 @@ function! codefmt#buildifier#GetFormatter() abort
     try
       " NOTE: Ignores any line ranges given and formats entire buffer.
       " buildifier does not support range formatting.
-      call codefmt#formatterhelpers#Format(l:cmd)
+      call codefmt#formatterhelpers#Format(maktaba#syscall#Create(l:cmd))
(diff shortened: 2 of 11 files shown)


autoload/codefmt/formatterhelpers.vim
```diff
@@ -0,0 +1,59 @@
+" Copyright 2017 Google Inc. All rights reserved.
```
malcolmr: ah, 2020? :)

autoload/codefmt/buildifier.vim
```diff
@@ -47,9 +47,9 @@ function! codefmt#buildifier#GetFormatter() abort
 
     let l:input = join(getline(1, line('$')), "\n")
```
malcolmr: Unused now.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -0,0 +1,59 @@
...
+" Unless required by applicable law or agreed to in writing, software
+" distributed under the License is distributed on an "AS IS" BASIS,
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+" See the License for the specific language governing permissions and
+" limitations under the License.
+
+
+""
+" @public
+" Format lines in the current buffer via a formatter invoked by {cmd}. The
```
malcolmr: Perhaps document that `{cmd}` is something you can pass to `maktaba#syscall#Create()`? As it stands, I wasn't sure if this was requiring a Syscall object, and the cases above where you passed a list were bugs.

(also below.)

autoload/codefmt/autopep8.vim
```diff
@@ -67,29 +67,18 @@ function! codefmt#autopep8#GetFormatter() abort
...
+          \ '--range', string(a:startline), string(a:endline),
+          \ '-']))
     else
-      let l:cmd = [l:executable, '-']
-      " Hack range formatting by formatting range individually, ignoring context.
-      let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
+      call codefmt#formatterhelpers#AttemptFakeRangeFormatting(
+          \ a:startline,
+          \ a:endline,
+          \ maktaba#syscall#Create([l:executable, '-']))
```
malcolmr: Any reason not to prefer just `[l:executable, '-']` here, and similarly `l:cmd` (etc) elsewhere, as you have for buildifier?

autoload/codefmt/formatterhelpers.vim
```diff
@@ -0,0 +1,59 @@
...
+" command includes the explicit range line numbers to use, if any.
+"
+" @throws ShellError if the {cmd} system call fails
+function! codefmt#formatterhelpers#Format(cmd) abort
```
malcolmr: Just wondering about how we might extend this for #60, if we wanted to centralize error-handling later.

Possibly we could add an extra optional argument that extracted the line:column:message from any error text, and caused this to handle adding errors to the quickfix list?

autoload/codefmt/formatterhelpers.vim
dbarnett: Fixed.

autoload/codefmt/buildifier.vim
dbarnett: Fixed.

autoload/codefmt/formatterhelpers.vim
dbarnett: Oh, good catch. It was a bug (but strangely wasn't caught by the tests… I'll need to investigate coverage there).

It actually does take a maktaba.Syscall… updated to clarify that and enforce.

autoload/codefmt/autopep8.vim
dbarnett: Yep, fair question. It'd save keystrokes but lose flexibility we'll want. For example, it wouldn't accommodate the `WithCwd` in zprint that I just rebased in.

autoload/codefmt/formatterhelpers.vim
dbarnett: AFAICT this accommodates it just fine if we provide another helper to accept the extracted errors (and let the caller handle the details of extracting the errors, which will vary case-by-case and won't generalize well).

autoload/codefmt/formatterhelpers.vim
malcolmr: I mean, I think there's some advantage in being able to pass in anything that can be passed to syscall#Create(), since then we can pass a list in the simple case, and a Syscall object in the uncommon case?

autoload/codefmt/formatterhelpers.vim
```diff
@@ -0,0 +1,74 @@
...
+" |maktaba.Syscall|). The command includes the explicit range line numbers to
+" use, if any.
+"
+" @throws ShellError if the {cmd} system call fails
+function! codefmt#formatterhelpers#Format(cmd) abort
+  call s:EnsureIsSyscall(a:cmd)
+  let l:lines = getline(1, line('$'))
+  let l:input = join(l:lines, "\n")
+
+  let l:result = maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call()
```
malcolmr: If a:cmd is a Syscall already, you don't need to call `Create()`.

(Though as mentioned elsewhere, maybe accepting either is fine? `Create(Syscall)` is documented to return itself, for convenience.)

autoload/codefmt/formatterhelpers.vim
dbarnett: Oh yeah! Good idea, I forgot we made it that clever. And I guess it didn't fail tests because it wasn't broken. :man_facepalming:

Updated to accept both.

autoload/codefmt/autopep8.vim
dbarnett: I gotcha now. Updated.

autoload/codefmt/formatterhelpers.vim
dbarnett: Resolved.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -0,0 +1,64 @@
...
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+" See the License for the specific language governing permissions and
+" limitations under the License.
+
+
+""
+" @public
+" Format lines in the current buffer via a formatter invoked by {cmd}, which
+" is a system call represented by either a |maktaba.Syscall| or any argument
+" accepted by |maktaba#syscall#Create()|. The command includes any arguments
```
malcolmr: "The command must include any", maybe?

dbarnett: Sure, why not? Done.

PR #136, opened 2020-01-29
feat(prettier): use npx for project-local prettier

agriffis:
If npx is available, use it for running prettier. This will find a project-local prettier, if available, which allows everybody on the team to run the same version. Otherwise people have different prettier versions, and code formatting toggles back and forth between commits.

Since npx can't tell us if prettier is available, we check with `npx prettier --version` and cache the result.

Also, since by default npx will do a temporary install of things you try to run, we use the `--no-install` flag to avoid that. We only want npx to find what's already there.

Along the way, kill the opinionated prettier options that were in vim-codefmt, since they unhelpfully override local and project config files.

### DIFF
--- a/autoload/codefmt/prettier.vim
+++ b/autoload/codefmt/prettier.vim
@@ -22,6 +22,14 @@ let s:supported_filetypes = ['javascript', 'markdown', 'html', 'css', 'yaml',
 
 ""
 " @private
+" Invalidates the cached prettier availability detection.
+function! codefmt#prettier#InvalidateIsAvailable() abort
+  unlet! s:prettier_is_available
+endfunction
+
+
+""
+" @private
 " Formatter: prettier
 function! codefmt#prettier#GetFormatter() abort
   let l:formatter = {
@@ -30,9 +38,21 @@ function! codefmt#prettier#GetFormatter() abort
           \ 'and configure the prettier_executable flag'}
 
   function l:formatter.IsAvailable() abort
-    let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray(
-          \ 'prettier_executable')
-    return !empty(l:cmd) && executable(l:cmd[0])
+    if !exists('s:prettier_is_available')
+      let s:prettier_is_available = 0
+      let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray(
+            \ 'prettier_executable')
+      if !empty(l:cmd) && executable(l:cmd[0])
+        " Unfortunately the availability of npx isn't enough to tell whether
+        " prettier is available, and npx doesn't have a way of telling us.
+        " Fetching the prettier version should suffice.
+        let l:result = maktaba#syscall#Create(l:cmd + ['--version']).Call(0)
+        if v:shell_error == 0
+          let s:prettier_is_available = 1
+        endif
+      endif
+    endif
+    return s:prettier_is_available
   endfunction
 
   function l:formatter.AppliesToBuffer() abort
@@ -52,7 +72,7 @@ function! codefmt#prettier#GetFormatter() abort
     if @% == ""
       call extend(l:cmd, ['--parser', 'babylon'])
     else
-      call extend(l:cmd, ['--stdin-filepath', @%])
+      call extend(l:cmd, ['--stdin-filepath', expand('%:p')])
     endif
 
     call maktaba#ensure#IsNumber(a:startline)
@@ -71,7 +91,13 @@ function! codefmt#prettier#GetFormatter() abort
           \ 'prettier_options'))
 
     try
-      let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
+      let l:syscall = maktaba#syscall#Create(l:cmd).WithStdin(l:input)
+      if isdirectory(expand('%:p:h'))
+        " Change to the containing directory so that npx will find
+        " a project-local prettier in node_modules
+        let l:syscall = l:syscall.WithCwd(expand('%:p:h'))
+      endif
+      let l:result = l:syscall.Call()
       let l:formatted = split(l:result.stdout, "\n")
       call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
     catch /ERROR(ShellError):/
--- a/instant/flags.vim
+++ b/instant/flags.vim
@@ -111,9 +111,7 @@ call s:plugin.Flag('shfmt_executable', 'shfmt')
 ""
 " Command line arguments to feed prettier. Either a list or callable that
 " takes no args and returns a list with command line arguments.
-call s:plugin.Flag('prettier_options', [
-    \ '--single-quote', '--trailing-comma=all',
-    \ '--arrow-parens=always', '--print-width=80'])
+call s:plugin.Flag('prettier_options', [])
 
 ""
 " The path to the prettier executable.
--- a/instant/flags.vim
+++ b/instant/flags.vim
@@ -114,8 +114,21 @@ call s:plugin.Flag('shfmt_executable', 'shfmt')
 call s:plugin.Flag('prettier_options', [])
 
 ""
-" The path to the prettier executable.
-call s:plugin.Flag('prettier_executable', 'prettier')
+" @private
+function s:LookupPrettierExecutable() abort
+  return executable('npx') ? ['npx', '--no-install', 'prettier'] : 'prettier'
+endfunction
+
+""
+" The path to the prettier executable. String, list, or callable that
+" takes no args and returns a string or a list. The default uses npx if
+" available, so that the repository-local prettier will have priority.
+call s:plugin.Flag('prettier_executable', function('s:LookupPrettierExecutable'))
+
+" Invalidate cache of detected prettier availability whenever
+" prettier_executable changes.
+call s:plugin.flags.prettier_executable.AddCallback(
+    \ maktaba#function#FromExpr('codefmt#prettier#InvalidateIsAvailable()'), 0)
 
 ""
 " Command line arguments to feed rustfmt. Either a list or callable that
--- a/doc/codefmt.txt
+++ b/doc/codefmt.txt
@@ -83,12 +83,13 @@ Default: 'shfmt' `
                                                     *codefmt:prettier_options*
 Command line arguments to feed prettier. Either a list or callable that takes
 no args and returns a list with command line arguments.
-Default: [ '--single-quote', '--trailing-comma=all', '--arrow-parens=always',
-  '--print-width=80'] `
+Default: [] `
 
                                                  *codefmt:prettier_executable*
-The path to the prettier executable.
-Default: 'prettier' `
+The path to the prettier executable. String, list, or callable that takes no
+args and returns a string or a list. The default uses npx if available, so
+that the repository-local prettier will have priority.
+Default: function('s:LookupPrettierExecutable') `
 
                                                      *codefmt:rustfmt_options*
 Command line arguments to feed rustfmt. Either a list or callable that takes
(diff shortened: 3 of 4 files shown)


agriffis:
@lhchavez @infokiller Looks like you've previously worked on prettier in vim-codefmt. I'd be interested in your thoughts on this.

infokiller:
The general idea makes sense and the code looks clean.
Why did you remove the opinionated prettier options that were not related to `textwidth`?

agriffis:
@infokiller Thanks for looking it over.

Regarding the opinionated options, I think they don't belong in vim-codefmt. Prettier already supplies defaults. The role of vim-codefmt is to bridge between Vim and the formatter, not to introduce a new set of defaults. Principle of least surprise.

However I think it makes sense, within reason, to propagate editor settings to formatters. As a user, I don't expect a formatter to run contrary to my Vim settings. For example, if I always write code at 100 columns width, I would like formatters to respect that instead of needing to configure them individually.

One of the nice outcomes is that it works with editorconfig. Per-project settings propagate seamlessly to formatters.

In fact, it would probably make sense for the prettier plugin to additionally respect `shiftwidth`, but that doesn't apply to zprint which is where I first implemented this concept. I might update this PR to add that.

Possibly this default policy should be settable by users. Perhaps there should be a global vim-codefmt setting that controls whether to inform formatters about Vim settings, so that users can easily change the policy in one place. What do you think?

infokiller:
Thanks for the great explanation. Sounds good to me, and I personally would also prefer vim to pass its settings to the formatters.
Regarding the opinionated prettier settings, I think they were introduced to make it behave more like clang-format, which IIRC was originally the only formatter for javascript in vim-codefmt.

dbarnett:
One note on inheriting formatting settings: it's not always an uncontroversial improvement to treat the buffer's style settings as the correct ones. We do see plenty of cases where repo owners configure opinionated local project style settings like .clang-format and that's supposed to trump textwidth, shiftwidth, etc., which devs sadly tend to ignore and expect things to "just work" when they save the file. There are also plugins like https://github.com/tpope/vim-sleuth that configure those settings with guesses and don't pay attention to local project configs like that. It would be great if something like editorconfig could clean up this mess for us, but do tread carefully on trying to change that behavior.

agriffis:
@infokiller @dbarnett Thanks to both of you for your thoughtful comments.

I just pushed a commit that refactors things and adds a new codefmt setting `apply_vim_settings`.

The idea here is that vim-codefmt can optionally apply buffer settings to formatters across the board, and it's easy for the user to disable this behavior globally for vim-codefmt.

I updated both the prettier and zprint plugins as a demonstration of using this, although that's a somewhat confusing example because zprint doesn't take normal command-line options, and I wanted `apply_vim_settings` to work reasonably even if `zprint_options` is also customized.

If you like where this is going, then I'll update the tests and also rebase into a few sensible commits rather than the existing stream-of-development.

agriffis:
Also, I made `apply_vim_setings` true by default because I think that makes sense out of the box, but making it false would be more conservative.

agriffis:
@dbarnett This is ready for proper review and pull when you're ready...

I rebased into four atomic commits, with tests passing on each commit. This pull-request now does two things:

1. Use the project-local prettier via npx if available.

2. Add the apply_vim_settings flag and use it in the prettier and zprint plugins. It defaults to zero, though, so nobody gets unpleasantly surprised. The downside is that people have to discover the flag, but that's the tradeoff.

I could do this as separate pull-requests, but they'd be sequential anyway, so I settled for atomic commits. Let me know what you think.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -62,3 +66,40 @@ function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
...
+  else
+    let l:flag = l:FlagFn
+  endif
+  if type(l:flag) is# type([])
+    return l:flag
+  elseif type(l:flag) is# type("") && l:flag !~ '\s'
+    return l:flag == "" ? [] : [l:flag]
+  endif
+  throw maktaba#error#WrongType(
+        \ a:flag_name . ' flag must be array after calling. Found %s',
```
dbarnett: This single error message doesn't clearly capture all the cases. For instance, it's not accurate to say if you pass a string containing a space that it "must be array after calling". I'd suggest splitting out a couple different error messages for different cases.

autoload/codefmt/zprint.vim
```diff
@@ -26,6 +26,30 @@
 
 let s:plugin = maktaba#plugin#Get('codefmt')
 
+""
+" @private
+function! s:ApplyToZprintOptions(opts, key, value) abort
```
dbarnett: Behavior here is a little hard to follow. It could use a doc comment and example input/output.

autoload/codefmt/zprint.vim
```diff
@@ -26,6 +26,30 @@
...
+  call maktaba#ensure#IsString(a:key)
+
+  " We can only apply Vim settings for simple values of a:opts, because it's
+  " an EDN map, not a list of command-line flags.
+  let l:edn = empty(a:opts) ? '{}' : a:opts[0]
+  if l:edn[0] != '{'
+    return a:opts
+  endif
+
+  " Don't try to insert if key already appears in the EDN.
```
dbarnett: It shouldn't try to override existing value for the key?

autoload/codefmt/zprint.vim
```diff
@@ -26,6 +26,30 @@
...
+  if l:edn =~ a:key
```
dbarnett: This pattern matching looks brittle. `=~` behavior depends on the `'ignorecase'` setting, and it'll match against values and parts of keys. Maybe you could split on whitespace and search through those strings, or something like that?

autoload/codefmt/prettier.vim
```diff
@@ -30,7 +30,9 @@ function! codefmt#prettier#GetFormatter() abort
           \ 'and configure the prettier_executable flag'}
 
   function l:formatter.IsAvailable() abort
-    return executable(s:plugin.Flag('prettier_executable'))
+    let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray(
```
dbarnett: Could you send a PR for setting up and using this new helper as a first step? I think it's a nice no-op refactoring which is easier to review if it's not combined with a new behavior change. Some of the diffs are a little harder to read when it's combined like this.

autoload/codefmt/zprint.vim
```diff
@@ -50,33 +76,33 @@ function! codefmt#zprint#GetFormatter() abort
...
-    else
-      throw maktaba#error#WrongType(
-            \ 'zprint_options flag must be list or callable. Found %s',
-            \ string(l:ZprintOptions))
+    let l:exe = codefmt#formatterhelpers#ResolveFlagToArray(
+          \ 'zprint_executable')
+    let l:opts = codefmt#formatterhelpers#ResolveFlagToArray(
+          \ 'zprint_options')
+
+    if s:plugin.Flag('apply_vim_settings')
```
dbarnett: A global flag won't support the per-path or per-language behavior this ought to support. Even if it's easy to disable it globally, you can now get into a position where you have a single case it doesn't work well for and you have this unhelpful choice between using it everywhere or using it nowhere.

dbarnett:
Sorry for the code review bomb... On the overall direction, I do like being able to hook the vim settings into the formatter's settings better, but I still have reservations about setting up a global `apply_vim_settings` flag that only "mostly" solves the cases I run into day-to-day.

I like the flag helper. Would you be up for sending a quick PR to add that first and streamline the scope of this PR?

agriffis:
Thanks @dbarnett. Busy at the moment but I'll get back to this soon

autoload/codefmt/formatterhelpers.vim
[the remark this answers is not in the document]
agriffis: done in #138

autoload/codefmt/formatterhelpers.vim
```diff
@@ -62,3 +66,40 @@ function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
...
+" Resolve a flag (function, string or array) to a normalized array, with special
+" handling to convert a spaceless string to a single-element array. This is the
+" common case for executables, and more importantly, is backward-compatible for
+" existing user settings.
+"
+" @throws WrongType if the flag doesn't resolve to a string or array
+function! codefmt#formatterhelpers#ResolveFlagToArray(flag_name) abort
+  let l:FlagFn = s:plugin.Flag(a:flag_name)
+  if maktaba#value#IsFuncref(l:FlagFn)
+    let l:flag = maktaba#function#Call(l:FlagFn)
```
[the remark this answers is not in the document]
agriffis: done in #138

autoload/codefmt/formatterhelpers.vim
```diff
@@ -62,3 +66,40 @@ function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
...
+  else
+    let l:flag = l:FlagFn
+  endif
+  if type(l:flag) is# type([])
+    return l:flag
+  elseif type(l:flag) is# type("") && l:flag !~ '\s'
```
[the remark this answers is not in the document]
agriffis: restructured and commented in #138

autoload/codefmt/formatterhelpers.vim
agriffis: done in #138

autoload/codefmt/prettier.vim
agriffis: Yep, see #138

autoload/codefmt/zprint.vim
agriffis: I'll plan to fix this in a separate PR

autoload/codefmt/zprint.vim
agriffis: You can't have duplicate keys in EDN, and replacing it would involve actually parsing the EDN, so I think it's better to defer in that case. But I'll handle this in a separate PR.

autoload/codefmt/zprint.vim
agriffis: Will address in a separate PR

autoload/codefmt/zprint.vim
agriffis: Will address in separate PR

agriffis:
@dbarnett I knocked this down to purely npx prettier and killing prettier options, as two separate commits.

All the stuff related to apply_vim_settings I'll move to a separate PR for reworking.

agriffis:
@dbarnett Let me know if you have any more questions about this

instant/flags.vim
```diff
@@ -111,13 +111,24 @@ call s:plugin.Flag('shfmt_executable', 'shfmt')
...
-call s:plugin.Flag('prettier_options', [
-    \ '--single-quote', '--trailing-comma=all',
-    \ '--arrow-parens=always', '--print-width=80'])
+call s:plugin.Flag('prettier_options', [])
 
 ""
-" The path to the prettier executable.
-call s:plugin.Flag('prettier_executable', 'prettier')
+" @private
+function s:PrettierExecutable() abort
```
dbarnett: Consider renaming this to "LookupPrettierExecutable" or something especially self-documenting like that since the name will appear in the output of `:Glaive codefmt`, and it'd take a little effort for the typical user who sees it to track down exactly what it's doing.

dbarnett:
PR description looks stale. Can you update it to clarify the latest state of the PR?

instant/flags.vim
agriffis: Done

agriffis:
> Can you run vimdoc to update the generated help file with your changes?

Done

> PR description looks stale. Can you update it to clarify the latest state of the PR?

Done

dbarnett:
Merged! Thanks for bearing with me on this one. =)

PR #138, opened 2020-02-25
refactor: add codefmt#formatterhelpers#ResolveFlagToArray

agriffis:
Refactor some common boilerplate so that it's easier to apply this logic
to executables and options across all plugins. Implemented in prettier
and zprint for starters, with no functional changes to either plugin.

Extracted from https://github.com/google/vim-codefmt/pull/136


dbarnett:
LGTM! Merging.

PR #139, opened 2020-02-25
fix(zprint): don't cd unless dir exists

agriffis:
This avoids the syscall failing if the directory doesn't exist yet. It's
an unlikely scenario, but I managed to hit it, and it's annoying when
tooling breaks in unlikely scenarios...


dbarnett:
+1, I too am a fanboy for correctness =)

PR #144, opened 2020-03-31
Ignore non-zero exit code from clang-format --version

### LINKED ISSUE #84: Error formatting file: ERROR(ShellError): Error running: clang-format --version
when I run `:Formatcode`, it turns out:
> Error formatting file: ERROR(ShellError): Error running: clang-format --version

```
[root@fe4524cdab25 ~]# clang-format -version
LLVM (http://llvm.org/):
  LLVM version 3.4.2
  Optimized build.
  Built Dec  7 2015 (09:37:36).
  Default target: x86_64-redhat-linux-gnu
  Host CPU: x86-64
```

dbarnett:
Also make version check logic a little more explicit and fix quirks for empty version or comparing shorter versions.

Resolves a ShellError bug when invoking some versions of the clang-format tool that return non-zero exit code, and instead makes it so if the tool returns a parseable version number it will be used regardless of exit code.

The material change was just updating `.Call()` to `.Call(0)`, but I refactored the version check logic a little in the process to make sure the scenario of empty version number is handled properly. Also adjusts the logic for comparing versions of mismatched lengths to pad version with zeros instead of checking length only, which could false negative on long min version values.

Fixes #84 .

### DIFF
--- a/autoload/codefmt/clangformat.vim
+++ b/autoload/codefmt/clangformat.vim
@@ -23,20 +23,35 @@ function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
       return 0
     endif
 
-    let l:version_output =
-          \ maktaba#syscall#Create([l:executable, '--version']).Call().stdout
+    let l:syscall = maktaba#syscall#Create([l:executable, '--version'])
+    " Call with throw_errors disabled because some versions of clang-format
+    " misbehave and return exit code 1 along with the successful version
+    " output (see https://github.com/google/vim-codefmt/issues/84).
+    let l:version_output = l:syscall.Call(0).stdout
     let l:version_string = matchstr(l:version_output, '\v\d+(.\d+)+')
+    " If no version string was matched, cached version will be an empty list.
     let s:clang_format_version = map(split(l:version_string, '\.'), 'v:val + 0')
   endif
-  let l:length = min([len(a:minimum_version), len(s:clang_format_version)])
+  " Always fail check if version couldn't be fetched.
+  if empty(s:clang_format_version)
+    return 0
+  endif
+  " Compare each dotted version value in turn.
+  let l:length = max([len(a:minimum_version), len(s:clang_format_version)])
   for i in range(l:length)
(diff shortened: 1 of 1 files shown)


autoload/codefmt/clangformat.vim
```diff
@@ -23,20 +23,35 @@ function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
...
+    " If no version string was matched, cached version will be an empty list.
     let s:clang_format_version = map(split(l:version_string, '\.'), 'v:val + 0')
   endif
-  let l:length = min([len(a:minimum_version), len(s:clang_format_version)])
+  " Always fail check if version couldn't be fetched.
+  if empty(s:clang_format_version)
+    return 0
+  endif
+  " Compare each dotted version value in turn.
+  let l:length = max([len(a:minimum_version), len(s:clang_format_version)])
```
agriffis, line 40: intentional change from min to max?

agriffis: ah, I see the change in the loop below, nevermind

autoload/codefmt/clangformat.vim
```diff
@@ -23,20 +23,35 @@ function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
...
+    if l:detected_value > l:minimum_value
       return 1
-    elseif a:minimum_version[i] > s:clang_format_version[i]
+    elseif l:detected_value < l:minimum_value
       return 0
     endif
   endfor
-  return len(a:minimum_version) <= len(s:clang_format_version)
+  " All version numbers were equal, so version was at least minimum.
+  return 1
```
agriffis, line 54: this version comparison code feels like something that would be useful as a general helper, eventually

dbarnett: For sure. We do have something similar in `maktaba#IsAtLeastVersion`, but without the variable length.

OTOH, often code that's checking versions is also low-level code that's awkward to have depending on libraries, but doesn't hurt to make an implementation available, at any rate.

PR #145, opened 2020-04-15
clang-format: work around vim/vim#5930 when positioning cursor

sam-mccall:
This bug (and a related predecessor) cause `:goto` and `line2bytes()`
to do the wrong thing when vim textprops are used.
This manifests as "cursor jumps around randomly after formatting" when using
plugins that use textprops for diagnostics, highlighting etc.

We happen to have all the code in an array when we need to do cursor coordinate
conversions, so we do those in vimscript instead of querying vim.

Without this patch, The newly added test fails on affected versions of vim
including both vim master and debian's 8.1.2269 (in different ways).

Vim bug tracker entry:
https://github.com/vim/vim/issues/5930
Similar workaround in first-party vim integration:
https://github.com/llvm/llvm-project/commit/591be7ec500c151d9232366042e21c74e006292c

### DIFF
--- a/autoload/codefmt/clangformat.vim
+++ b/autoload/codefmt/clangformat.vim
@@ -120,32 +120,49 @@ function! codefmt#clangformat#GetFormatter() abort
       let l:cmd += ['-lines', l:startline . ':' . l:endline]
     endfor
 
+    let l:lines = getline(1, line('$'))
+
     " Version 3.4 introduced support for cursor tracking
     " http://llvm.org/releases/3.4/tools/clang/docs/ClangFormat.html
     let l:supports_cursor = s:ClangFormatHasAtLeastVersion([3, 4])
     if l:supports_cursor
-      " line2byte counts bytes from 1, and col counts from 1, so -2 
-      let l:cursor_pos = line2byte(line('.')) + col('.') - 2
+      " Avoid line2byte: https://github.com/vim/vim/issues/5930
+      let l:cursor_pos = col('.') - 1
+      for l:i in range(1, line('.') - 1)
+        let l:cursor_pos += len(l:lines[l:i - 1]) + 1
+      endfor
       let l:cmd += ['-cursor', string(l:cursor_pos)]
     endif
 
-    let l:input = join(getline(1, line('$')), "\n")
+    let l:input = join(l:lines, "\n")
     let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
     let l:formatted = split(l:result.stdout, "\n")
 
-    if !l:supports_cursor
-      call maktaba#buffer#Overwrite(1, line('$'), l:formatted[0:])
-    else
-      call maktaba#buffer#Overwrite(1, line('$'), l:formatted[1:])
+    if l:supports_cursor
+      " With -cursor, the first line is a JSON object.
+      let l:header = l:formatted[0]
+      let l:formatted = l:formatted[1:]
+      call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
       try
-        let l:clang_format_output_json = maktaba#json#Parse(l:formatted[0])
-        let l:new_cursor_pos =
-            \ maktaba#ensure#IsNumber(l:clang_format_output_json.Cursor) + 1
-        execute 'goto' l:new_cursor_pos
+        let l:header_json = maktaba#json#Parse(l:header)
+        let l:offset = maktaba#ensure#IsNumber(l:header_json.Cursor)
+        " Compute line/col, avoid goto: https://github.com/vim/vim/issues/5930
+        let l:new_line = 0
+        for l:line in l:formatted
+          let l:offset_after_next = l:offset - len(l:line) - 1
+          if l:offset_after_next < 0
+            break
+          endif
+          let l:offset = l:offset_after_next
+          let l:new_line += 1
+        endfor
+        call cursor(l:new_line + 1, l:offset + 1)
       catch
         call maktaba#error#Warn('Unable to parse clang-format cursor pos: %s',
             \ v:exception)
       endtry
+    else
+      call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
     endif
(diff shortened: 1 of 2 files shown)


autoload/codefmt/clangformat.vim
```diff
@@ -120,32 +120,49 @@ function! codefmt#clangformat#GetFormatter() abort
...
     " http://llvm.org/releases/3.4/tools/clang/docs/ClangFormat.html
     let l:supports_cursor = s:ClangFormatHasAtLeastVersion([3, 4])
     if l:supports_cursor
-      " line2byte counts bytes from 1, and col counts from 1, so -2 
-      let l:cursor_pos = line2byte(line('.')) + col('.') - 2
+      " Avoid line2byte: https://github.com/vim/vim/issues/5930
+      let l:cursor_pos = col('.') - 1
+      for l:i in range(1, line('.') - 1)
+        let l:cursor_pos += len(l:lines[l:i - 1]) + 1
+      endfor
```
dbarnett, line 133: Consider refactoring this into an `s:Func`. Later it might make a nice helper in itself in `maktaba#buffer#`.

autoload/codefmt/clangformat.vim
```diff
@@ -120,32 +120,49 @@ function! codefmt#clangformat#GetFormatter() abort
...
     let l:formatted = split(l:result.stdout, "\n")
 
-    if !l:supports_cursor
-      call maktaba#buffer#Overwrite(1, line('$'), l:formatted[0:])
-    else
-      call maktaba#buffer#Overwrite(1, line('$'), l:formatted[1:])
+    if l:supports_cursor
+      " With -cursor, the first line is a JSON object.
+      let l:header = l:formatted[0]
+      let l:formatted = l:formatted[1:]
```
dbarnett, line 144: Could shorten these two lines to
```viml
let l:header = remove(l:formatted, 0)
```

autoload/codefmt/clangformat.vim
```diff
@@ -120,32 +120,49 @@ function! codefmt#clangformat#GetFormatter() abort
...
+      call maktaba#buffer#Overwrite(1, line('$'), l:formatted)
       try
-        let l:clang_format_output_json = maktaba#json#Parse(l:formatted[0])
-        let l:new_cursor_pos =
-            \ maktaba#ensure#IsNumber(l:clang_format_output_json.Cursor) + 1
-        execute 'goto' l:new_cursor_pos
+        let l:header_json = maktaba#json#Parse(l:header)
+        let l:offset = maktaba#ensure#IsNumber(l:header_json.Cursor)
+        " Compute line/col, avoid goto: https://github.com/vim/vim/issues/5930
```
dbarnett, line 149: Here too, this should be an `s:Func` for now for clarity and potentially added as a maktaba helper.

dbarnett:
Thanks for digging into this and fixing!

autoload/codefmt/clangformat.vim
sam-mccall: Sure thing! I'm not sure how useful these are apart from as a workaround for this bug, I'll let you make the call on moving them.

sam-mccall:
Thanks for the review, and sorry about the hacky workarounds!

autoload/codefmt/clangformat.vim
```diff
@@ -55,6 +55,36 @@ function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
 endfunction
 
 
+" Inputs are 1-based (row, col) coordinates into lines.
+" Returns the corresponding zero-based offset into lines->join("\n")
+function! s:PositionToOffset(row, col, lines) abort
+  let l:offset = a:col - 1
+  if a:row > 1
+    for l:line in a:lines[0 : a:row - 2]
+      let l:offset += len(l:line) + 1
```
dbarnett, line 64: The "+1" is for the newline, right? Could you comment one or two inline hints explaining the arbitrary-looking +N and -N parts?

autoload/codefmt/clangformat.vim
```diff
@@ -55,6 +55,36 @@ function! s:ClangFormatHasAtLeastVersion(minimum_version) abort
...
+" Returns the 1-based [row, col] coordinates into lines.
+function! s:OffsetToPosition(offset, lines) abort
+  let l:lines_consumed = 0
+  let l:chars_left = a:offset
+  for l:line in a:lines
+    let l:chars_after_next = l:chars_left - len(l:line) - 1
+    if l:chars_after_next < 0
+      break
+    endif
+    let l:chars_left = l:chars_after_next
```
dbarnett, line 81: Up to you, but here's an equivalent that's slightly easier to read for me, flipping the `l:chars_after_next` math:
```suggestion
    let l:line_len = len(l:line) + 1
    if l:chars_left < l:line_len
      break
    endif
    let l:chars_left -= l:line_len
```

autoload/codefmt/clangformat.vim
sam-mccall: Done. The -2 is particularly awful.

(I wonder if there'd be enough usage of a maktaba#list#slice(begin, end) that returned a half-open range... `[begin, end)` - the `if` and extra `-1` to deal with closed ranges is horrible, and I don't know a better idiom)

sam-mccall:
@dbarnett Sorry for the long turnaround, and thanks for the review - this should be good to go.

vroom/clangformat.vroom
```diff
@@ -178,7 +178,33 @@ that clang-format has a version >= 3.4).
...
 
+This should work when with textprops set in the file, despite various functions
+like line2byte() and :goto being buggy in that case.
+(See https://github.com/vim/vim/issues/5930 for bug details)
 
+  @clear
+  % int f()     {<CR>
+  |    int i=1;<CR>
+  |   return 1234567890; }<CR>
+  :call prop_type_add('keyword', {})
```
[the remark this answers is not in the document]
sam-mccall: Oops, sorry about that. Sent #150

PR #147, opened 2020-04-26
associate arduino file-type with clang-format

yayayaty:
Arduino (https://www.arduino.cc/) source files (.ino filename extension) are essentially a c++ files, but recognized as a separate file type by vim (for Arduino-specific syntax highlighting, I believe). 
Arduino IDE, which uses .ino files for user source code, provides for external editor support, which could be vim.
I found it helpful to associate arduino files with clang-format, which works well for me.


yayayaty:
@googlebot I signed it!

PR #149, opened 2020-05-08
Fix default parser for Prettier (babylon -> babel)

brentyi:
Minor fix that only impacts unnamed buffers: `babylon` was deprecated in v1.16.0 (prettier/prettier#5647) and removed in v2.0.0 (prettier/prettier#7533).


dbarnett:
Thanks!

dbarnett:
Looks like this never passed the tests in Travis: https://travis-ci.org/github/google/vim-codefmt/builds/684632452. Weird that it didn't notify the PR with failures...

PR #150, opened 2020-05-18
clang-format: fix tests on old vim without textprop

sam-mccall:
d8dd427b9edd0c71d9373eb2696730d45d5bad75 broke the CI: https://travis-ci.org/github/google/vim-codefmt/jobs/680186570.

### DIFF
--- a/vroom/clangformat.vroom
+++ b/vroom/clangformat.vroom
@@ -186,8 +186,10 @@ like line2byte() and :goto being buggy in that case.
   % int f()     {<CR>
   |    int i=1;<CR>
   |   return 1234567890; }<CR>
-  :call prop_type_add('keyword', {})
-  :call prop_add(1, 1, {'length': 3, 'type': 'keyword'})
+  :if has('textprop')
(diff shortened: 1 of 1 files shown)


vroom/clangformat.vroom
```diff
@@ -186,8 +186,10 @@ like line2byte() and :goto being buggy in that case.
   % int f()     {<CR>
   |    int i=1;<CR>
   |   return 1234567890; }<CR>
-  :call prop_type_add('keyword', {})
-  :call prop_add(1, 1, {'length': 3, 'type': 'keyword'})
+  :if has('textprop')
```
dbarnett, line 189: Nit: Add a note to clarify we're not verifying anything interesting in the `!has('textprop')` case. Something like this:
```suggestion
This skips the textprop case if not supported (meaning this block won't do any
interesting verification when run in older versions of vim).
  :if has('textprop')
```

dbarnett:
Obsoleted by #159.

PR #152, opened 2020-06-07
Add support for arbitrary shfmt command with arguments

funous:
Make it possible to use complex shfmt commands, such as shfmt running in a docker container. Use the codefmt#formatterhelpers#ResolveFlagToArray approach seen in other formatters.

Similar to https://github.com/google/vim-codefmt/pull/124

### DIFF
--- a/doc/codefmt.txt
+++ b/doc/codefmt.txt
@@ -82,7 +82,8 @@ Google's style. See https://github.com/mvdan/sh for details.
 Default: ['-i', '2', '-sr', '-ci'] `
 
                                                     *codefmt:shfmt_executable*
-The path to the shfmt executable.
+The path to the shfmt executable. String, list, or callable that takes no args
+and returns a string or a list.
 Default: 'shfmt' `
 
                                                     *codefmt:prettier_options*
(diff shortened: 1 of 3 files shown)


doc/codefmt.txt
```diff
@@ -82,7 +82,8 @@ Google's style. See https://github.com/mvdan/sh for details.
 Default: ['-i', '2', '-sr', '-ci'] `
 
                                                     *codefmt:shfmt_executable*
-The path to the shfmt executable.
+The path to the shfmt executable. String, list, or callable that takes no args
+and returns a string or a list.
```
dbarnett, line 86: Looks like you manually edited this without updating the source in flags.vim it's generated from, so it'll get reverted next time someone regenerates the help docs.

funous: I've updated the flags.vim and checked that vimdoc generates the same string. I've also updated flags.vim to have the source for the changes I've done manually in https://github.com/google/vim-codefmt/pull/124. I can split it into another PR if it's preferred.

dbarnett: Thanks! Same PR is fine for this, no worries.

PR #154, opened 2020-07-07
Add OCaml support with ocamlformat

adamyi:


adamyi:
It seems that the failed tests are unrelated to this cl, but in prettier and clang-format. The ocamlformat vroom tests passed :)

autoload/codefmt/ocamlformat.vim
```diff
@@ -0,0 +1,86 @@
...
+      " See https://github.com/ocaml-ppx/ocamlformat/pull/1188
+      call codefmt#formatterhelpers#AttemptFakeRangeFormatting(
+          \ a:startline, a:endline, l:cmd)
+    catch /ERROR(ShellError):/
+      " Parse all the errors and stick them in the quickfix list.
+      let l:errors = []
+      let l:matchidx = 1
+      while 1
+        let l:tokens = matchlist(v:exception,
+            \ '\vFile ' . l:fname_pattern . ', line (\d+), characters (\d+)-\d+:\n(.*)\n', 0, l:matchidx)
```
dbarnett, line 63: Instead of the double indirection matching the output using (1) a pattern built from (2) a dynamic variable, could you just do a looser match for anything that looks like "filename or <standard input>" and skip any that don't match? This seems like it could be brittle if e.g. paths may be absolute/relative, or have any trivial extra characters like "./".

PR #158, opened 2020-07-28
IsAvailable: Fix "type mismatch" and print nicer errors

### LINKED ISSUE #155: when I use formatcode clang-format it shows such errors
error detected while processing function <SNR>38_FormatBufferAndSetRepeat..codefmt#FormatBuffer..<SNR>43_GetFormatter ..<SNR>43_IsAvailable..12..codefmt#formatterhelpers#ResolveFlagToArray:

dbarnett:
Prints user-friendly errors if a formatter fails an IsAvailable call (also allowing other formatters to still be checked), and fixes a particular "type mismatch" error encountered on vim versions older than
7.4.1546.

Fixes #155.

### DIFF
--- a/autoload/codefmt.vim
+++ b/autoload/codefmt.vim
@@ -107,20 +107,40 @@ endfunction
 " @function(#SetWhetherToPerformIsAvailableChecksForTesting), skips the
 " IsAvailable check and always returns true.
 function! s:IsAvailable(formatter) abort
-  if codefmt#ShouldPerformIsAvailableChecks()
-    return a:formatter.IsAvailable()
+  if !codefmt#ShouldPerformIsAvailableChecks()
+    return 1
   endif
-  return 1
+  return a:formatter.IsAvailable()
+endfunction
+
+
+" Checks whether {formatter} is available, safely handling errors by logging
+" an error and returning 0.
+function! s:IsAvailableSafe(formatter) abort
+  try
+    return s:IsAvailable(a:formatter)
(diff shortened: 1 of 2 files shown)


agriffis:
LGTM

Kashomon:
Postfacto LGTM =)

autoload/codefmt.vim
```diff
@@ -107,20 +107,40 @@ endfunction
...
 ""
 " Detects whether a formatter has been defined for the current buffer/filetype.
 function! codefmt#IsFormatterAvailable() abort
-  let l:formatters = copy(s:registry.GetExtensions())
-  let l:is_available = 'v:val.AppliesToBuffer() && s:IsAvailable(v:val)'
-  return !empty(filter(l:formatters, l:is_available)) ||
-      \ !empty(get(b:, 'codefmt_formatter'))
+  if !empty(get(b:, 'codefmt_formatter'))
+    return 1
+  endfor
```
dbarnett, line 137: Wut. I somehow used `:endfor` instead of `:endif` here without seeing any test failures. Gotta fix that and check what's up with test coverage…

dbarnett: Oh, because the tests usually disable it with `#SetWhetherToPerformIsAvailableChecksForTesting()`. :facepalm: 

Stay tuned for fix.

dbarnett: Fixed in 9346c381c.

PR #163, opened 2020-09-13
Add support for cmake-format.

antmusco:
Includes `autoload/codefmt/cmakeformat.vim` which adds support for `cmake-format` for formatting CMake files. Adds two additional flags, `cmake_format_executable`, which defaults to `cmake-format`, and `cmake_format_config`, which defaults to the empty string (i.e. default formatter).


autoload/codefmt/cmakeformat.vim
```diff
@@ -0,0 +1,78 @@
...
+    let l:fname = expand('%:p')
+    if empty(l:fname)
+        return
+    endif
+    let l:cmd += [l:fname]
+
+    " Generate formatted output.
+    let l:input = join(getline(1, line('$')), "\n")
+    let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
+    let l:formatted = split(l:result.stdout, "\n")
```
dbarnett, line 71: I think you want to use `split`'s keepempty argument here, `split(STDOUT, "\n", 1)`, so you have consistent results if the first or last line is empty.

autoload/codefmt/cmakeformat.vim
```diff
@@ -0,0 +1,78 @@
...
+
+    " Overwrite buffer.
+    call maktaba#buffer#Overwrite(1, line('$'), l:formatted[0:])
```
dbarnett, line 74: What's the `[0:]` slicing for? Should be a no-op except for making a copy of the list before passing it to Overwrite, which AFAIK isn't needed.

instant/flags.vim
```diff
@@ -66,6 +66,17 @@ call s:plugin.flags.clang_format_executable.AddCallback(
...
 call s:plugin.Flag('clang_format_style', 'file')
 
+""
+" The path to the cmake-format executable. String, list, or callable that
+" takes no args and returns a string or a list.
+call s:plugin.Flag('cmake_format_executable', 'cmake-format')
+
+""
+" The path to the cmake-format configuration file for cmake-format to use.
+" See https://cmake-format.readthedocs.io/en/latest/installation.html for
```
dbarnett, line 76: Looks like actually https://cmake-format.readthedocs.io/en/latest/configuration.html has the details and installation.html doesn't say much about it.

instant/flags.vim
```diff
@@ -66,6 +66,17 @@ call s:plugin.flags.clang_format_executable.AddCallback(
...
+" details.
+call s:plugin.Flag('cmake_format_config', '')
```
dbarnett, line 78: The flag is global, so it will be a single value used for every file you edit. Is that what you'd want?

By my reading of https://cmake-format.readthedocs.io/en/latest/configuration.html, you can pass multiple config files, and that's independent of the automatic detection of files like `\.?cmake-format(.yaml|.json|.py)` that it finds on the path? Can you confirm if an explicit config overrides any automatic config detection and note that in the doc comments?

dbarnett (changes requested):
If you're able to run the vimdoc tool from http://github.com/google/vimdoc to generate help files from your comments, that would be helpful.

morxa:
@antmusco I'd really like to have support for cmake-format, so if you could pick this up again, I'd be very grateful :) Please let me know if I can help!

dbarnett:
@morxa looks like this got abandoned. If you'd like to pick up the changes and send your own PR we can finish the review and try to get something merged.

PR #165, opened 2020-11-10
Treat any yapf exit code as an error case, not just exit code 1

### LINKED ISSUE #61: yapf code formatting deletes file if it does not parse

dbarnett:
Fixes #61.


malcolmr (approved):
LGTM

PR #167, opened 2020-12-13
Add formatter for lua 5.1

ElPiloto:
luaformatterfiveone is a formatter for Lua 5.1.  It is readily available
through standard via `luarocks install formatterfiveone`.

PR #170, opened 2021-03-27
js-beautify support for mustache and handlebars

abgeana:
The mustache/vim-mustache-handlebars plugin sets the filetype to either
html.handlebars or html.mustache. Even though js-beautify supports
formatting for these files, the `--type` parameter must be `html`.

This commit sets the `--type` parameter to `html` when the filetype
matches the regex `html\..*`.

### DIFF
--- a/autoload/codefmt/jsbeautify.vim
+++ b/autoload/codefmt/jsbeautify.vim
@@ -30,7 +30,8 @@ function! codefmt#jsbeautify#GetFormatter() abort
   endfunction
 
   function l:formatter.AppliesToBuffer() abort
-    return &filetype is# 'css' || &filetype is# 'html' || &filetype is# 'json' ||
+    return &filetype is# 'css' || &filetype is# 'html' ||
+        \ &filetype =~# '\mhtml\.' || &filetype is# 'json' ||
         \ &filetype is# 'javascript'
   endfunction
 
@@ -45,6 +46,8 @@ function! codefmt#jsbeautify#GetFormatter() abort
       let l:cmd = l:cmd + ['--type', 'js']
     elseif &filetype is# 'sass' || &filetype is# 'scss' || &filetype is# 'less'
       let l:cmd = l:cmd + ['--type', 'css']
+    elseif &filetype =~# '\mhtml\.'
+      let l:cmd = l:cmd + ['--type', 'html']
     elseif &filetype != ""
       let l:cmd = l:cmd + ['--type', &filetype]
     endif


abgeana:
@googlebot I signed it!

abgeana:
CI is still failing, but I see in the tests for js-beautify are fine this time. The problem is with another formatter I think.

autoload/codefmt/jsbeautify.vim
```diff
@@ -30,7 +30,7 @@ function! codefmt#jsbeautify#GetFormatter() abort
   endfunction
 
   function l:formatter.AppliesToBuffer() abort
-    return &filetype is# 'css' || &filetype is# 'html' || &filetype is# 'json' ||
+    return &filetype is# 'css' || &filetype =~ 'html.*' || &filetype is# 'json' ||
         \ &filetype is# 'javascript'
```
dbarnett, line 34: This loose matching includes some patterns you might not have intended. Consider changing it to something like
```vim
&filetype is# 'html' || &filetype =~# '\mhtml\.'
```

Without the `#` the behavior subtly depends on the `'ignorecase'` setting, and the loose `.*` will match on characters that aren't a literal dot, such as some weird `htmlos` filetype vim has that doesn't seem to be anything like actual HTML syntax.

autoload/codefmt/jsbeautify.vim
```diff
@@ -45,8 +45,11 @@ function! codefmt#jsbeautify#GetFormatter() abort
       let l:cmd = l:cmd + ['--type', 'js']
     elseif &filetype is# 'sass' || &filetype is# 'scss' || &filetype is# 'less'
       let l:cmd = l:cmd + ['--type', 'css']
+    elseif &filetype =~ 'html\..*'
+      let l:cmd = l:cmd + ['--type', 'html']
     elseif &filetype != ""
       let l:cmd = l:cmd + ['--type', &filetype]
+    elseif &filetype
```
dbarnett, line 52: Was this line here by accident? It's "harmless" but has no effect.

dbarnett:
> CI is still failing, but I see in the tests for js-beautify are fine this time. The problem is with another formatter I think.

Interesting, looks like #167 and for some reason Travis never ran on that. I'll look into it.

dbarnett (changes requested):

abgeana:
I addressed the two changes you requested and also squashed the commits to make history nicer. Let me know if this works.

dbarnett:
Thanks, merging!

dbarnett (approved):

PR #173, opened 2021-03-28
Consolidate jsbeautify supported formats and allow for sass/scss/less

dbarnett:
Deduplicates per-filetype logic in AppliesToBuffer & FormatRange to use a shared `_supported_formats` definition. In the process, updates AppliesToBuffer to accept sass/scss/less filetypes, where #114 had only fixed them to work via explicit `:FormatCode jsbeautify` but not via `:FormatCode`.

@abgeana @awused FYI


malcolmr (approved):
LGTM

PR #177, opened 2021-04-15
Tests: Fix several failures related to formatter precedence

dbarnett:
Fixes several vroom tests that wrongly assume their formatter will be the default for a given filetype and started failing when I reordered the formatter registrations in b1b51098.

Note those precedence changes could really affect users, but only if they're relying on defaults and have multiple installed formatters for those same filetypes. If so, it's easily resolved by explicitly specifying a formatter name or uninstalling unwanted formatters.

### DIFF
--- a/vroom/autopep8.vroom
+++ b/vroom/autopep8.vroom
@@ -60,18 +60,6 @@ You can format any buffer with autopep8 specifying the formatter explicitly.
       pass
   @end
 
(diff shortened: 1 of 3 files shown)


dbarnett:
I don't think there's any "correct" precedence among these default formatters in general, but LMK if anyone does have any precedence you think is important to guarantee.

PR #189, opened 2022-01-19
Don't call clang-format on empty buffer

ezzieyguywuf:
- FIXME: this test doesn't actually do anything yet
- Do not call clang-format if buffer is empty.

I did my best to generate a unit test that duplicates the bug, however I've stared
it at so long I'm going cross-eyed and I still cannot figure out a good way to do this.

My problem is I'm not sure how to check the output of `:FormatCode clang-format` for
an error (or lack thereof). I tried using a `try-catch` but was unsure how to
do this with vroom while still intercepting the system calls to provide a mocked
response.

Any help on figuring out a good unit test for this would be greatly appreciated.

### DIFF
--- a/autoload/codefmt/clangformat.vim
+++ b/autoload/codefmt/clangformat.vim
@@ -129,6 +129,11 @@ function! codefmt#clangformat#GetFormatter() abort
   ""
   " Reformat buffer with clang-format, only targeting [ranges] if given.
   function l:formatter.FormatRanges(ranges) abort
+    if  line('$') == 1 && getline(1) == ''
+      " If the buffer is completely empty, there's nothing to do
+      return
+    endif
+
     let l:Style_value = s:plugin.Flag('clang_format_style')
     if type(l:Style_value) is# type('')
       let l:style = l:Style_value
--- a/vroom/clangformat.vroom
+++ b/vroom/clangformat.vroom
@@ -181,3 +181,10 @@ current buffer.
 
 The -assume-filename arg passed above is also related, used by the clang-format
 tool to detect any custom style rules particular to the current file.
+
+
+Bug #102. When the buffer is empty, clang-format does _not_ return a cursor
+position
+
+  @clear
+  :FormatCode clang-format


dbarnett:
I think you want `maktaba#error#TryCommand`: https://github.com/google/vim-maktaba/blob/46730b0d818da2da005e3c8a38ff987a2dd36d7c/vroom/error.vroom#L265

Vroom is easily confused by splitting stateful commands across lines and even more confused by trying to catch errors in vimscript.

autoload/codefmt/clangformat.vim
```diff
@@ -182,8 +182,8 @@ function! codefmt#clangformat#GetFormatter() abort
         let [l:new_line, l:new_col] = s:OffsetToPosition(l:offset, l:formatted)
         call cursor(l:new_line, l:new_col)
       catch
-        call maktaba#error#Warn('Unable to parse clang-format cursor pos: %s',
-            \ v:exception)
+        call maktaba#error#Warn(v:exception)
```
dbarnett, line 185: What's the motivation for changing the warning? Ideally the context is printed before the exception message. Maybe separating with a newline instead of a space would help with what you're going for?

dbarnett (changes requested):

autoload/codefmt/clangformat.vim
dbarnett: Oh, I believe that's more a convention for error messages *thrown* from library functions and actually used sort of accidentally with "Warn" there, where Warn is usually human-readable informational messages directed at users.

It sounds like the "regexing for unexpected errors" you're talking about is the behavior from the default ["GUESS-ERRORS" option](https://github.com/google/vroom/blob/abd0e4ad98254e7c245c58ce22bd1635179494f0/vroom/args.py#L237) for message strictness. That only has any effect for messages you don't otherwise match as expected (see [example in messages.vroom](https://github.com/google/vroom/blob/abd0e4ad98254e7c245c58ce22bd1635179494f0/examples/messages.vroom#L14)). You should be able to control that however you need without modifying the warning logging, but can you clarify what you're aiming to get the test to do?

ezzieyguywuf: > can you clarify what you're aiming to get the test to do?

I'm trying to write a test that duplicates the bug.

The main approach I'm using is

1. Call `FormatCode`
2. intercept the system call and return a blank string
3. Make sure there aren't any errors

As it stands, since the code emits a warning, I don't have a very robust way of ensuring that `FormatCode` runs without any problems.

I _could_ check for the specific warning output, but even this doesn't help since the warning is only displayed when the error occurs. Once the error is fixed, there is no output, and I was unable to find a way to check for "the function call returns no output"

I can try the `--message=strict` thing, I must have missed that the handful of times I read through that document

dbarnett: Oh, gotcha! You want to temporarily change message strictness to `STRICT` with the `@messages` directive ([example](https://github.com/google/vroom/blob/abd0e4ad98254e7c245c58ce22bd1635179494f0/examples/directives.vroom#L28-L45)).

ezzieyguywuf: That's exactly what I needed :-D

Thanks for the help. Marking this as resolved now, though let me know if you have any other comments.

vroom/clangformat.vroom
```diff
@@ -181,3 +181,14 @@ current buffer.
...
 The -assume-filename arg passed above is also related, used by the clang-format
 tool to detect any custom style rules particular to the current file.
+
+
+Bug #102. When the buffer is empty, clang-format does _not_ return a cursor
+position
+
+  @clear
+  @messages (STRICT)
+  :call maktaba#error#TryCommand('FormatCode clang-format')
```
dbarnett, line 191: Does it actually throw errors in the error case, or just warn? It might actually work fine if you just directly invoke `:FormatCode clang-format`, but fine to keep it this way if there is a reason that gets weird.

autoload/codefmt/clangformat.vim
```diff
@@ -171,7 +171,7 @@ function! codefmt#clangformat#GetFormatter() abort
     let l:result = maktaba#syscall#Create(l:cmd).WithStdin(l:input).Call()
     let l:formatted = split(l:result.stdout, "\n")
 
-    if l:supports_cursor
+    if l:supports_cursor && !(line('$') == 1 && getline(1) == '')
```
dbarnett, line 174: Would it make sense to do the empty buffer check above and avoid sending `-cursor` at all? That way it's not depending on subtle behaviors that might change.

dbarnett (changes requested):
Looks good, one question about the clang-format invocation and one minor question in the vroom test.

vroom/clangformat.vroom
ezzieyguywuf: Yup, this is simply a carryover from me desperately trying to write a unit test that actually did what I wanted :-P

I've updated the test as you suggested.

autoload/codefmt/clangformat.vim
ezzieyguywuf: Ideally I'd prefer to bail out of the function as early as possible if the buffer is empty - really there's nothing to do.

Unfortunately there's a few other tests that rely on `clang-format` getting called no matter what when `FormatCode` is called.

I considered fixing the tests, but that was expanding the scope a bit too much for the amount of time I was willing to invest.

I could open a bug if you'd like so that at least we can keep track of it.

I guess another option is to call `clang-format` at the beginning of the function and then bail out if the buffer is empty.

dbarnett: Right, skipping the whole formatting call would do similar, but would break more tests and intuitive assumptions. My main concern is relying on subtle (undocumented?) behavior that could conceivably change out from under us and have really strange results, if clang-format did someday decide to start returning the requested cursor coords in that case for consistency and we misparsed them.

If it's not too many affected tests and we can fix them in this PR to not have empty buffers, that'd be ideal (I don't think it's _that_ many tests that explicitly check for the `-cursor` option?). Otherwise, a tracking bug is better than nothing.

ezzieyguywuf: I'll take a look, you're probably right that it's probably not _that_ many tests.

ezzieyguywuf: 😅

turns out that moving the check to the beginning doesn't affect _any_ other tests. I don't know why I thought it did - I thought before when I tried it, I saw some broken tests. but I just ran `vroom vroom/` and didn't get any failures.

marking this as resolved.

vroom/clangformat.vroom
```diff
@@ -181,3 +181,10 @@ current buffer.
 
 The -assume-filename arg passed above is also related, used by the clang-format
 tool to detect any custom style rules particular to the current file.
+
+
+Bug #102. When the buffer is empty, clang-format does _not_ return a cursor
+position
+
+  @clear
+  :FormatCode clang-format
```
dbarnett, line 190: Did you not want to use `@messages (STRICT)` here to catch `Unable to parse clang-format cursor pos`? Or did you find the test already catches the problem as-is if you run it without the fix?

ezzieyguywuf: this test catches the problem as-is when run without the fix.

dbarnett (approved):

PR #192, opened 2022-05-20
Add --lint_mode flag for buildifier

TamaMcGlinn:
In your vimrc you can do e.g.

```
call glaive#Install()
Glaive codefmt buildifier_lint_mode='fix'
```

to get buildifier to autofix issues. The default behaviour is unchanged.

### DIFF
--- a/autoload/codefmt/buildifier.vim
+++ b/autoload/codefmt/buildifier.vim
@@ -39,7 +39,11 @@ function! codefmt#buildifier#GetFormatter() abort
   " @flag(buildifier)
   " @throws ShellError
   function l:formatter.Format() abort
+    let l:lint_flag = s:plugin.Flag('buildifier_lint_mode')
     let l:cmd = [ s:plugin.Flag('buildifier_executable') ]
+    if l:lint_flag != ""
+      let l:cmd += ["--lint=" . l:lint_flag]
+    endif
     let l:fname = expand('%:p')
     if !empty(l:fname)
       let l:cmd += ['-path', l:fname]
--- a/doc/codefmt.txt
+++ b/doc/codefmt.txt
@@ -105,6 +105,14 @@ Default: 'gn' `
 The path to the buildifier executable.
 Default: 'buildifier' `
 
+                                               *codefmt:buildifier_lint_mode*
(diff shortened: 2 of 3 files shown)


dbarnett:
Wouldn't it be better to by default not pass any mode override at all? This would be another layer of defaulting that could fight with other local config settings, etc.

TamaMcGlinn:
I'm passing the default value though, so that will do exactly the same. Anyway, I like being explicit about the lint_mode. If this could fight with other local config settings, that means you're saying the lint_mode can already be specified in some other way, and the whole PR would be unnecessary. As far as I know, that's not the case, but if so, please elaborate on how to do that.

dbarnett:
I don't know whether buildifier would have any other override mechanism in practice or not, just that I prefer not to copy-paste defaults. IOW, you're saying it will do the same because you're passing the same default value… what if they change the default later, like add a new middle option for minimal fixes and make that the default? Would we want this default to take precedence over theirs?

TamaMcGlinn:
Okay, fixed that. Also tested and specified in the help message what the behaviour is for 'warn', since it is possible someone would actually choose that behaviour.

autoload/codefmt/buildifier.vim
```diff
@@ -39,7 +39,11 @@ function! codefmt#buildifier#GetFormatter() abort
   " @flag(buildifier)
   " @throws ShellError
   function l:formatter.Format() abort
+    let l:lint_flag = s:plugin.Flag('buildifier_lint_mode')
     let l:cmd = [ s:plugin.Flag('buildifier_executable') ]
+    if l:lint_flag != ""
+      let l:cmd += ["--lint=" .. l:lint_flag]
```
dbarnett, line 45: Is this `..` a typo intended to be `.`? AFAICT it blows up with an error on my version of vim.

dbarnett:
Thanks for working with me on the defaults! Looks good now besides one `..` thing that might be broken I commented on.

autoload/codefmt/buildifier.vim
TamaMcGlinn: How strange! I'm using NeoVim and always used two dots to concatenate strings, I never knew one dot sufficed. It's fixed now.

dbarnett:
Merged, thanks!

PR #197, opened 2022-07-11
Add --warnings flag for buildifier

TamaMcGlinn:
In your vimrc you can do e.g.

```
call glaive#Install()
Glaive codefmt buildifier_lint_mode='fix'
Glaive codefmt buildifier_warnings='-module-docstring,+unsorted-dict-items'
```

to get buildifier to ignore missing module docstrings but sort
dictionaries automatically. The default behaviour is unchanged.

### DIFF
--- a/autoload/codefmt/buildifier.vim
+++ b/autoload/codefmt/buildifier.vim
@@ -41,8 +41,12 @@ function! codefmt#buildifier#GetFormatter() abort
   function l:formatter.Format() abort
     let l:lint_flag = s:plugin.Flag('buildifier_lint_mode')
     let l:cmd = [ s:plugin.Flag('buildifier_executable') ]
-    if l:lint_flag != ""
+    if !empty(l:lint_flag)
       let l:cmd += ["--lint=" . l:lint_flag]
+    endif
+    let l:warnings_flag = s:plugin.Flag('buildifier_warnings')
+    if !empty(l:warnings_flag)
+      let l:cmd += ["--warnings=" . l:warnings_flag]
     endif
     let l:fname = expand('%:p')
     if !empty(l:fname)
--- a/doc/codefmt.txt
+++ b/doc/codefmt.txt
@@ -117,6 +117,23 @@ Options:
   an error and do no formatting.
 Default: '' `
 
+                                                *codefmt:buildifier_warnings*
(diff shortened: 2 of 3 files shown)


autoload/codefmt/buildifier.vim
```diff
@@ -44,6 +44,10 @@ function! codefmt#buildifier#GetFormatter() abort
     if l:lint_flag != ""
       let l:cmd += ["--lint=" . l:lint_flag]
     endif
+    let l:warnings_flag = s:plugin.Flag('buildifier_warnings')
+    if l:warnings_flag != ""
+      let l:cmd += ["--warnings=" . l:warnings_flag]
```
dbarnett, line 49: I *think* if there are spaces in the flag value this will do the right thing since `maktaba#syscall` will do escaping, but could you verify?

autoload/codefmt/buildifier.vim
```diff
@@ -44,6 +44,10 @@ function! codefmt#buildifier#GetFormatter() abort
     if l:lint_flag != ""
       let l:cmd += ["--lint=" . l:lint_flag]
     endif
+    let l:warnings_flag = s:plugin.Flag('buildifier_warnings')
+    if l:warnings_flag != ""
```
dbarnett, line 48: Consider using `!empty(l:warnings_flag)` since vim's equality checking has some bizarre corner cases (like `:echo 0 == 'walrusface'` --> `true`).

autoload/codefmt/buildifier.vim
TamaMcGlinn: I'm not sure what you mean by 'the right thing', since spaces are not allowed in this flag. I could add a check and throw an error in that case, but then the question arises; why not check all other invalid characters as well?

I tried this:

```
Glaive codefmt buildifier_lint_mode='' buildifier_warnings='+unsorted-dict-items --lint=invalid'
```

and it just ignores the invalid flag - so I guess it must be inserting a backslash before the space and therefore not passing --line=invalid as a separate argument. If we later add a flag that is allowed to contain spaces, this should do the right thing.

autoload/codefmt/buildifier.vim
TamaMcGlinn: done

dbarnett:
Thanks! Merging.

PR #201, opened 2022-10-20
Add mixformat to handle Elixir

### LINKED ISSUE #200: Formatter for Elixir
**Formatter tool**
`mix` is the standard Elixir build tool.  It provides a [`mix format`](https://hexdocs.pm/mix/main/Mix.Tasks.Format.html) command.

**Filetype(s)**
`elixir` (`*.ex`, `*.exs`), `eelixir` (`*.eex`), `heex` (`*.heex`)

**Additional context**
`mix format` can accept a file via stdin, but doesn't seem to support line range formatting.  It derives formatting configuration from a `.formatter.exs` config file which can be at the root of a project (parallel to `mix.exs`) or overridden by a formatter in a subdirectory.  It also works on Elixir files which are not part of a Mix project.

flwyd:
Fixes #200.


autoload/codefmt/mixformat.vim
```diff
@@ -0,0 +1,108 @@
...
+let s:cmdAvailable = {}
+
+""
+" @private
+" Formatter: mixformat
+function! codefmt#mixformat#GetFormatter() abort
+  let l:formatter = {
+      \ 'name': 'mixformat',
+      \ 'setup_instructions': 'mix is usually installed with Elixir' .
+          \ '(https://elixir-lang.org/install.html). ' .
```
dbarnett, line 25: Did you mean for this to be "with Elixir (https://...)" here with a space after "Elixir", or was "with Elixir(https://...)" intentional?

autoload/codefmt/mixformat.vim
```diff
@@ -0,0 +1,108 @@
...
+  function l:formatter.FormatRange(startline, endline) abort
+    let l:filename = expand('%:p')
+    if empty(l:filename)
+      let l:filename = 'stdin.exs'
+    endif
+    " mix format docs: https://hexdocs.pm/mix/main/Mix.Tasks.Format.html
+    let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray('mix_executable')
+    " Specify stdin as the file
+    let l:cmd = l:cmd + ['format', '--stdin-filename=' . l:filename, '-']
+    let l:dir = l:filename == 'stdin.exs' ? getcwd() : s:findMixDir(l:filename)
```
dbarnett, line 53: Can stdin.exs be a valid filename here, or is it just supposed to be a placeholder value? If so, would this give the right behavior for a file explicitly named stdin.exs?

dbarnett (changes requested):

autoload/codefmt/mixformat.vim
flwyd: Good catch.

autoload/codefmt/mixformat.vim
flwyd: Fair point.  I moved l:dir assignment into the `if empty` block and added a comment giving context to `stdin.exs`.

dbarnett (approved):

dbarnett:
Thanks, Trevor!

PR #205, opened 2022-11-13
Add ocamlformat formatter

### LINKED ISSUE #204: Formatter for OCaml
**Formatter tool**
[ocamlformat](https://github.com/ocaml-ppx/ocamlformat)

**Filetype(s)**
OCaml

**Additional context**
`ocamlformat` does not seem to support range formatting. It reads formatting settings from `.ocamlformat` file, which it searches for in input file directory, going up into parent directores until it finds one. When formatting stdin, one of `--name`, `--impl`, or `--intf` options has to be provided.

dimitrijer:
Fixes #204.

### DIFF
--- a/autoload/codefmt/ocamlformat.vim
+++ b/autoload/codefmt/ocamlformat.vim
@@ -1,0 +1,60 @@
+" Copyright 2021 Google Inc. All rights reserved.
+"
+" Licensed under the Apache License, Version 2.0 (the "License");
+" you may not use this file except in compliance with the License.
+" You may obtain a copy of the License at
+"
+"     http://www.apache.org/licenses/LICENSE-2.0
+"
+" Unless required by applicable law or agreed to in writing, software
+" distributed under the License is distributed on an "AS IS" BASIS,
+" WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


autoload/codefmt/ocamlformat.vim
```diff
@@ -0,0 +1,60 @@
...
+    let l:executable = s:plugin.Flag('ocamlformat_executable')
+
+    " ocamlformat requires --name, --impl, or --intf when reading from
+    " stdin.
+    let l:inputflags = ['--name', @%]
+    if len(@%) == 0
+      " Assume we're formatting an implementation file.
+      let l:inputflags = ['--impl']
+    endif
+    call codefmt#formatterhelpers#Format(
```
dbarnett, line 53: If this hits some error in formatting, it'd blow up with an ugly error. For now could you add a TODO to catch and handle errors more nicely?

dbarnett (changes requested):
One question about error handling. Looks good otherwise.

autoload/codefmt/ocamlformat.vim
dimitrijer: Good catch, done.

dbarnett:
Thanks!

PR #206, opened 2022-12-01
register jsonnetfmt using buildifier as a template

### LINKED ISSUE #156: Add a formatter for jsonnet
**Formatter tool**
jsonnetfmt https://github.com/google/go-jsonnet/tree/master/cmd/jsonnetfmt

**Filetype(s)**
`.jsonnet` and `.libsonet` files.

**Additional context**
Doc for jsonnet: https://jsonnet.org/

nakulj:
Fixes #156


autoload/codefmt/jsonnetfmt.vim
```diff
@@ -0,0 +1,46 @@
...
+  let l:formatter = {
+      \ 'name': 'jsonnetfmt',
+      \ 'setup_instructions': 'Install jsonnet. ' .
+          \ '(https://jsonnet.org/).'}
+
+  function l:formatter.IsAvailable() abort
+    return executable(s:plugin.Flag('jsonnetfmt_executable'))
+  endfunction
+
+  let s:supported_filetypes = ['json','jsonnet']
```
dbarnett, line 18: Did you mean `l:supported_filetypes` instead of `s:`? Kind of weird to define a script-global variable that's only needed locally right below the definition.

autoload/codefmt/jsonnetfmt.vim
```diff
@@ -0,0 +1,46 @@
...
+    if !empty(l:fname)
+      let l:cmd += ['-path', l:fname]
+    endif
+
+    try
+      " NOTE: Ignores any line ranges given and formats entire buffer.
+      " jsonnetfmt does not support range formatting.
+      call codefmt#formatterhelpers#Format(l:cmd)
+    catch
+      " Parse all the errors and stick them in the quickfix list.
```
dbarnett, line 40: Was this supposed to say `TODO: Parse all ...`? It doesn't look like the code does what the comment says.

dbarnett (changes requested):
Two questions, looks good to me otherwise.

autoload/codefmt/jsonnetfmt.vim
nakulj: To be honest, I just copied this bit from https://github.com/google/vim-codefmt/blob/master/autoload/codefmt/prettier.vim :) I'll change it.

autoload/codefmt/jsonnetfmt.vim
nakulj: Updated the comment. Again, this was mostly copied from buildifier.vim

dbarnett (approved):
Thanks! Merging.

yegle:
Hmm the jsonnetfmt (both C++ version and Go version) does not have a `-path` flag. Was this a mistake?

dbarnett:
Yeah, probably copypasta from some other formatter file? That means the formatter always fails if you try to format a buffer with an associated filename?

Easy enough to remove if so.

dbarnett:
I filed #231 for that and have a fix I'm about to submit for it.

PR #208, opened 2022-12-04
update jsonnetfmt command

nakulj:
By default jsonnetfmt outputs to stdout. Use it in-place instead.


dbarnett:
Do you mean to directly edit the file on disk instead of formatting the code in the vim buffer? That tends to confuse vim if so.

dbarnett:
Fixed properly in #232 (needs to operate on stdin, not on a file path).

PR #209, opened 2023-02-09
Allow commands to be followed by a `|`.

chaoren:
This is useful for allowing simple conditional formatting of buffers, e.g.,

```vim
autocmd FileType * if condition | AutoFormatBuffer | endif
```


dbarnett:
Seems reasonable, thanks for the improvement!

Merging.

PR #219, opened 2023-05-25
Fix broken ktfmt availability check

### LINKED ISSUE #218: `ktfmt` availability check is broken (on Linux)
**Describe the bug**
When trying to use `ktfmt` I get the following error:

```
Formatter "ktfmt" is not available. Setup instructions: Install ktfmt (https://github.com/facebookincubator/ktfmt). Enable with                                                                                                                                                          
Glaive codefmt ktfmt_executable=java,-jar,/path/to/ktfmt-<VERSION>-jar-with-dependencies.jar   
```

However, if I reverse the conditional check [like this](https://github.com/grodtron/vim-codefmt/commit/55c6d76deba107852ffcef771f76abaadf468cc0) then everything works fine. I feel like I may be missing something since I would have thought others would encounter this issue, but I have tested this locally and it fixes the issue for me. If you agree that this change makes sense I can sign the CLA and make a pull request, or feel free to just make the (one character) change directly.

**To Reproduce**
Minimal vimrc snippet, does not include installing codefmt itself:

```
Glaive codefmt                                                                                                                                                                                                                                                                           
    \ plugin[mappings]=',='                                                                                                                                                                                                                                                              
    \ ktfmt_executable=java,-jar,/local/home/grodtron/.tools/ktfmt-0.44-jar-with-dependencies.jar                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                         
" Auto format on buffer write                                                                                                                                                                                                                                                            
augroup autoformat                                                                                                                                                                                                                                                                       
    autocmd!                                                                                                                                                                                                                                                                             
    autocmd FileType kotlin AutoFormatBuffer ktfmt                                                                                                                                                                                                                                       
augroup END                           
```

1. Open any kotlin file
2. Write the file (`:w`)

**Expected behavior**
The file should be formatted

**OS (version)**
Linux

**Additional context**
n/a

grodtron:
(Fixes issue #218) When trying to use this locally, I was getting repeated errors that ktfmt was not available. Reversing the conditional here makes everything work correctly for me (on Linux), and also makes sense (success of a command line tool is indicated by return code equal to 0).

### DIFF
--- a/autoload/codefmt/ktfmt.vim
+++ b/autoload/codefmt/ktfmt.vim
@@ -45,7 +45,7 @@ function! codefmt#ktfmt#GetFormatter() abort
         let l:success = 0
         try
           let l:result = maktaba#syscall#Create(l:cmd + ['-']).Call()
-          let l:success = v:shell_error != 0
+          let l:success = v:shell_error == 0
         catch /ERROR(ShellError)/
           call maktaba#error#Warn(
                 \ 'ktfmt unavailable, check jar file in `%s -`: %s',


dbarnett:
LGTM.

Per discussion in the PR, it already gives decent warnings if it does fail, so if anyone does start seeing issues they can give us a nice actionable report and we can fix corner cases pretty quickly.

dbarnett (approved):

PR #132, opened 2019-12-30
Add support for rubocop formatter

zinovyev:
This pull request adds a rubocop support for Ruby and ERuby file types.
Solves https://github.com/google/vim-codefmt/issues/30

Issue #30, opened 2015-02-21
Add a formatter for ruby


fowles:
https://github.com/erniebrodeur/ruby-beautify

dbarnett:
Someone also suggested rubocop in http://www.reddit.com/r/vim/comments/2uab1g/codefmt_a_plugin_for_formatting_code_using/co9pn5e. Wonder how it compares to ruby-beautify…

fowles:
The --auto-correct option is marked experimental for robocop.

https://github.com/bbatsov/rubocop

zinovyev:
I think that rubocop is somehow like an industry standard nowadays.

PR #220, opened 2023-05-27
Add ruby formatting support using Rubocop.

akveritas:
Solves #30 

This change is based on an abandoned change by zinovyev: 
https://github.com/zinovyev/vim-codefmt/tree/add-support-for-ruby
and matching pull request:
https://github.com/google/vim-codefmt/pull/132

I've added code comments to resolve dbarnett's question from that review, and added vroom tests.

### DIFF
--- a/autoload/codefmt/formatterhelpers.vim
+++ b/autoload/codefmt/formatterhelpers.vim
@@ -93,7 +93,7 @@ endfunction
 " N lines of the formatter output are trimmed. This can be used to trim
 " always-present headers.
 "
-" @throws ShellError if the {cmd} system call fails (and vararg0 is not set)
+" @throws ShellError if the {cmd} system call fails (and ignoreerrors is 0)
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
     \ startline, endline, cmd, ignoreerrors = 0, skipfirstnlines = 0) abort
--- a/autoload/codefmt/formatterhelpers.vim
+++ b/autoload/codefmt/formatterhelpers.vim
@@ -75,6 +75,7 @@ endfunction
 
 ""
 " @public
+" @usage startline endline cmd \[ignoreerrors] \[skipfirstnlines]
 " Attempt to format a range of lines from {startline} to {endline} in the
 " current buffer via a formatter that doesn't natively support range
 " formatting, which is invoked via {cmd} (a system call represented by either
@@ -93,24 +94,28 @@ endfunction
 " N lines of the formatter output are trimmed. This can be used to trim
 " always-present headers.
 "
-" @throws ShellError if the {cmd} system call fails (and ignoreerrors is 0)
+" @throws ShellError if the {cmd} system call fails (and [ignoreerrors] is 0)
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
-    \ startline, endline, cmd, ignoreerrors = 0, skipfirstnlines = 0) abort
+    \ startline, endline, cmd, ...) abort
   call maktaba#ensure#IsNumber(a:startline)
   call maktaba#ensure#IsNumber(a:endline)
-  call maktaba#ensure#IsNumber(a:ignoreerrors)
-  call maktaba#ensure#IsNumber(a:skipfirstnlines)
+
+  let l:ignoreerrors = a:0 >= 1 ? a:1 : 0
+  let l:skipfirstnlines = a:0 >= 2 ? a:2 : 0
+
+  call maktaba#ensure#IsNumber(l:ignoreerrors)
+  call maktaba#ensure#IsNumber(l:skipfirstnlines)
 
   let l:lines = getline(1, line('$'))
   let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
 
   let l:result =
-      \ maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call(!a:ignoreerrors)
+      \ maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call(!l:ignoreerrors)
   let l:formatted = split(l:result.stdout, "\n")
   " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
   let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
-  let l:full_formatted = l:before + l:formatted[a:skipfirstnlines :]
+  let l:full_formatted = l:before + l:formatted[l:skipfirstnlines :]
       \ + l:lines[a:endline :]
 
   call maktaba#buffer#Overwrite(1, line('$'), l:full_formatted)
--- a/autoload/codefmt/formatterhelpers.vim
+++ b/autoload/codefmt/formatterhelpers.vim
@@ -90,18 +90,21 @@ endfunction
 " @throws ShellError if the {cmd} system call fails
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
-    \ startline, endline, cmd) abort
+    \ startline, endline, cmd, ignoreerrors, skipfirstnlines) abort
   call maktaba#ensure#IsNumber(a:startline)
   call maktaba#ensure#IsNumber(a:endline)
 
   let l:lines = getline(1, line('$'))
   let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
 
-  let l:result = maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call()
+  let l:result = a:ignoreerrors ?
+      \ maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call(0)
+      \ : maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call() 
   let l:formatted = split(l:result.stdout, "\n")
   " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
   let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
-  let l:full_formatted = l:before + l:formatted + l:lines[a:endline :]
+  let l:full_formatted = l:before + l:formatted[a:skipfirstnlines :]
+      \ + l:lines[a:endline :]
 
   call maktaba#buffer#Overwrite(1, line('$'), l:full_formatted)
 endfunction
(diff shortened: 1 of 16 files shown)


autoload/codefmt/rubocop.vim
```diff
@@ -0,0 +1,70 @@
...
+  " Reformat the current buffer with rubocop or the binary named in
+  " @flag(rubocop_executable), only targeting the range between {startline} and
+  " {endline}.
+  " @throws ShellError
+  function l:formatter.FormatRange(startline, endline) abort
+    " See flag explanations at:
+    " https://docs.rubocop.org/rubocop/1.51/usage/basic_usage.html
+    let l:cmd = [s:plugin.Flag('rubocop_executable'), '--stdin', @%, '-a', '--no-color', '-fq', '-o', '/dev/null']
+
+    call maktaba#ensure#IsNumber(a:startline)
```
dbarnett, line 45: Where this is basically a copy-paste of parts of `codefmt#formatterhelpers#AttemptFakeRangeFormatting`, do you think for maintainability's sake it would be possible to factor this out as a named helper like `s:AttemptFakeRangeFormattingIgnoringFirstLine` to at least try to keep the common logic as one-to-one with AttemptFakeRangeFormatting as possible (and eventually consolidate those somehow)?

vroom/rubocop.vroom
```diff
@@ -0,0 +1,59 @@
...
+  |  call add(g:repeat_calls, a:000)<CR>
+  :endfunction
+  :call maktaba#test#Override('repeat#set', 'FakeRepeat')
+
+  :call codefmt#SetWhetherToPerformIsAvailableChecksForTesting(0)
+
+By default, the rubocop executable is called. To use this plugin, rubocop
+must be installed on your system.
+  :FormatCode rubocop
+  ! rubocop .*
```
dbarnett, line 19: Won't this fail for someone who doesn't have rubocop installed?

I could be wrong but I believe if you match a shell call like this but don't hijack to stub out any stdout or anything it'll still try to call the external command on your system. Prefer to stub out expected rubocop behavior so
  (a) if that behavior changes out from under us and breaks the plugin it's easier to understand what happened, and
  (b) we can run all the plugin's tests without needing to have dozens of different external tools installed.

dbarnett: If you can uninstall rubocop and still get the test passing I'm probably just mistaken about how the shell hijacks work. It doesn't seem to choke on the some_other_program part like I'd expect (or maybe it does but the failure is too silent and doesn't bomb out of the whole test run?).

akveritas: I'm new to vroom, but my understanding is that system calls are all stubbed out by default. You can see it doing what you're asking down on line 26: it makes a call to a non-existent program. If you add command output, with $, and expectations, the dummy command works like all the other calls. I kept the test case minimal just to make it clearer what each case is testing.

autoload/codefmt/rubocop.vim
akveritas: I stared at this for a few minutes, and I don't think we should pull this out as a separate function, because the error handling in this case is unusual. We can't just use Call() in the function and catch the exception, because that throws away the result. But I'd expect most callers to want normal error handling.

I could generalize AttemptFakeRangeFormatting with two new parameters, e.g. ignorefirstnlines and ignorefailures, if you think it's worth the complexity for other callers. Then you end up with something like:


function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
...
  let l:result = ignorefailures ? maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call(0) : maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call()
  " Special case empty slice: neither l:lines[:0] nor l:lines[:-1] is right.
  let l:before = a:startline > 1 ? l:lines[ : a:startline - 2] : []
  let l:full_formatted = l:before + l:formatted[ignorefirstnlines :] + l:lines[a:endline :]
...

akveritas: ... I went ahead and did this; let me know which version you prefer.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -90,18 +90,21 @@ endfunction
 " @throws ShellError if the {cmd} system call fails
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
-    \ startline, endline, cmd) abort
+    \ startline, endline, cmd, ignoreerrors, skipfirstnlines) abort
```
dbarnett, line 93: Can you make them optional using varargs, since they're kind of obscure options and almost always 0, 0?

autoload/codefmt/formatterhelpers.vim
```diff
@@ -90,18 +90,21 @@ endfunction
...
   call maktaba#ensure#IsNumber(a:startline)
   call maktaba#ensure#IsNumber(a:endline)
 
   let l:lines = getline(1, line('$'))
   let l:input = join(l:lines[a:startline - 1 : a:endline - 1], "\n")
 
-  let l:result = maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call()
+  let l:result = a:ignoreerrors ?
+      \ maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call(0)
+      \ : maktaba#syscall#Create(a:cmd).WithStdin(l:input).Call()
```
dbarnett, line 102: Instead of this big ternary you can just pass in !a:ignoreerrors at the end: `let l:result = maktaba#syscall#Create(a:cmd).WithStdint(l:input).Call(!a:ignoreerrors)`.

vroom/rubocop.vroom
dbarnett: Alright, I had a closer look and it seems legit.

autoload/codefmt/formatterhelpers.vim
akveritas: Facepalm. Done.

autoload/codefmt/formatterhelpers.vim
akveritas: Done.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -87,21 +87,32 @@ endfunction
 " the tool for range formatting and post a URL for that feature request above
 " code that calls it.
 "
-" @throws ShellError if the {cmd} system call fails
+" vararg0: if truthy, the syscall ignores errors. This can be helpful for
```
dbarnett, line 90: The convention is to put the actual name in square braces (instead of calling it vararg0) and vimdoc detects it as an optional parameter.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -87,21 +87,32 @@ endfunction
...
+"
+" @throws ShellError if the {cmd} system call fails (and vararg0 is not set)
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
-    \ startline, endline, cmd) abort
+    \ startline, endline, cmd, ...) abort
   call maktaba#ensure#IsNumber(a:startline)
   call maktaba#ensure#IsNumber(a:endline)
 
+  let l:ignoreerrors = a:0 >= 1 ? a:1 : 0
```
dbarnett, line 103: Might still want to explicitly verify types on these if they're passed.

autoload/codefmt/formatterhelpers.vim
akveritas: In the process of hunting for documentation on vimdoc, I found out that vimscript can use default values. That seems more appropriate than variable args here.

I reformatted the comment to be hopefully more idiomatic; PTAL.

autoload/codefmt/formatterhelpers.vim
akveritas: Done.

autoload/codefmt/formatterhelpers.vim
dbarnett: As of 8.1.1310, apparently: https://github.com/vim/vim/commit/42ae78cfff

I'm sure it's paranoia, the code was submitted in 2019, but still we tend to err on the side of supporting very old vim versions. I'm trying to cross-reference when that version would've reached e.g. Debian stable and other major distros... Would you be able to help research that for my peace of mind? 😅

dbarnett: Also still need to add support for it in the vimdoc tool apparently: https://github.com/google/vimdoc/issues/114

Does it work if you explicitly add `@default ignoreerrors 0` in the doc comments, or does vimdoc just complain that it's not optional and shouldn't have a default?

akveritas: Is there a command somewhere to generate the vimdoc? Or can you explain what you mean by "does it work"? I'm new to vim development so I'm not sure what I need to test.

akveritas: OK, looking at Ubuntu LTS releases, 20.04 Focal has vim 8.1.2269 (OK), but 18.04 Bionic has vim 8.0.1453 (Not OK). Bionic is just over the five year mark; 14.04 is still getting security updates so in theory someone could be using this plugin there. I miss Live at Head from my Google days.

I went ahead and reverted to the vararg version, and gave vimdoc a @usage hint. PTAL.

autoload/codefmt/formatterhelpers.vim
```diff
@@ -87,21 +88,35 @@ endfunction
...
+"
+" @throws ShellError if the {cmd} system call fails (and [ignoreerrors] is 0)
 " @throws WrongType
 function! codefmt#formatterhelpers#AttemptFakeRangeFormatting(
-    \ startline, endline, cmd) abort
+    \ startline, endline, cmd, ...) abort
   call maktaba#ensure#IsNumber(a:startline)
   call maktaba#ensure#IsNumber(a:endline)
 
+  let l:ignoreerrors = a:0 >= 1 ? a:1 : 0
```
dbarnett, line 104: Fun fact: You can also do `let l:ignoreerrors = get(a:, 1, 0)` for the same effect, slightly more cryptic but a bit more literal once you're used to the syntax.

Either way works fine here though, and I'm focused on getting this PR merged quickly after all the delays. =)

autoload/codefmt/formatterhelpers.vim
dbarnett: Thanks for humoring me there. 🤓

dbarnett: > Is there a command somewhere to generate the vimdoc?

And yes, there's a [vimdoc](https://github.com/google/vimdoc) tool to fully generate these docs. I can just tweak and regen those on my end after we merge this for simplicity's sake.

dbarnett (approved):

PR #225, opened 2023-08-15
Fix dartfmt's IsAvailable() method.

### LINKED ISSUE #190: dartfmt is deprecated
**Describe the bug**
The `dart format` command replaces `dartfmt`.
Dart SDK version: 2.16.1 already removed `dartfmt` command line tool. 
I try to use alias as a work around, but not working. 
alias dartfmt='dart format -o show' 

thanks!

malcolmr:
d927656 changed the `dartfmt_executable` flag from a string to a list, but didn't change the use inside IsAvailable(), leading to:

```
Failed to evaluate whether formatter dartfmt is available: Vim(return):E730: Using a List as a String
```

The change uses the `ResolveFlagToArray()` helper in both `IsAvailable()` and `FormatRange()` (which incidentally also allows defining the flag using a Function), and updates the documentation for all flags using `ResolveFlagToArray()` to use the same phrasing.

Fixes #190.

### DIFF
--- a/autoload/codefmt/dartfmt.vim
+++ b/autoload/codefmt/dartfmt.vim
@@ -26,7 +26,13 @@ function! codefmt#dartfmt#GetFormatter() abort
           \ 'https://dart.dev/get-dart'}
 
   function l:formatter.IsAvailable() abort
-    return executable(s:plugin.Flag('dartfmt_executable'))
+    let l:cmd = codefmt#formatterhelpers#ResolveFlagToArray(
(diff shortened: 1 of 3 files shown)


malcolmr:
I'm not sure why the neovim tests are failing, but it doesn't seem to be due to this change. Unfortunately that looks like that failure also aborts the Vim tests, but those do pass locally.

dbarnett (approved):

Issue #210, opened 2023-02-17
Formatter for Perl


liuyehcf:
**Formatter tool**
[Name of the tool and link to context]

[perltidy](https://perltidy.sourceforge.net/)

**Filetype(s)**
[Applicable languages/filetypes you'd like to use this tool to format]

* perl

**Additional context**
[Add any other context or screenshots about the feature request here.]

PR #227, opened 2023-08-20
Add perl formatting support using Perltidy

liuyehcf:
Solves #210

### DIFF
--- a/vroom/perltidy.vroom
+++ b/vroom/perltidy.vroom
@@ -1,0 +1,38 @@
+Perltidy is a linter and formatter for perl.
+If you aren't familiar with basic codefmt usage yet, see main.vroom first.
+
+First, set up the vroom environment.
+
+  :source $VROOMDIR/setupvroom.vim
+
+  :let g:repeat_calls = []
+  :function FakeRepeat(...)<CR>
+  |  call add(g:repeat_calls, a:000)<CR>
+  :endfunction
+  :call maktaba#test#Override('repeat#set', 'FakeRepeat')
+
+  :call codefmt#SetWhetherToPerformIsAvailableChecksForTesting(0)
+
+By default, the perltidy executable is called. To use this plugin, perltidy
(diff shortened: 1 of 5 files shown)


vroom/perltidy.vroom
```diff
@@ -0,0 +1,38 @@
...
+  :endfunction
+  :call maktaba#test#Override('repeat#set', 'FakeRepeat')
+
+  :call codefmt#SetWhetherToPerformIsAvailableChecksForTesting(0)
+
+By default, the perltidy executable is called. To use this plugin, perltidy
+must be installed on your system. (But not for testing; vroom intercepts
+system calls.)
+  :FormatCode perltidy
+  ! perltidy .*
```
dbarnett, line 20: You need to fake the command output like this: https://github.com/google/vim-codefmt/blob/c6730fb9c0d17060ea0caf5b9bd3090e86f6a14d/vroom/clangformat.vroom#L46

Otherwise the tests can't run without perltidy installed (and may get brittle, failing depending on the _version_ of perltidy installed).

dbarnett:
Sorry, I think since the tests weren't passing this fell through the cracks. I think if you can rebase your changes on top of the latest upstream and address my comment there to fix the tests, this looks ready to merge otherwise.

dbarnett (changes requested):

vroom/perltidy.vroom
liuyehcf: Updated, and I'm not sure if I'm doing right because I don't really understand the syntax of `.vroom`.

dbarnett:
I can help update the tests if a current maintainer can review. @malcolmr @fowles @snu5mumr1k

dbarnett:
Here ya go, patch https://github.com/dbarnett/vim-codefmt/commit/de65871fd6b926e3f8fca74dabe0e1a2e4285087 and you should be good to go.

PR #230, opened 2023-11-15
Add Julia formatter

flwyd:
This formatter uses a script wrapping the [JuliaFormatter.jl library](https://github.com/domluna/JuliaFormatter.jl).  This "formatter script in the plugin" approach hasn't been taken before in `vim-codefmt`, so let me know if you'd prefer a different approach or to reorganize things.  (The JuliaFormatter.jl library ships with an executable script, but it only works on file paths (not stdin) and can't handle line ranges.  I'm also not sure how to get the path of a bin directory in your Julia environment, which doesn't work like executables in a package management system like npm.)

If the user hasn't installed the `JuliaFormatter` package, instructions will point to an `install` script in the same directory as the formatter script.  I've tested the shell version of this script, but don't have a Windows machine to test the `.cmd` version.

The line-range handling is a bit of a hack: the underlying library doesn't expose this functionality, so I used comments to enable/disable the formatter around relevant blocks.  I found some edge case behaviors that I've hopefully documented clearly.

The script will use the same `.JuliaFormatter.toml` files in the parent directory hierarchy of the source file being edited to determine the formatter options like style, indent, and line width.  This is the same behavior that running the executable from JuliaFormatter.

### DIFF
--- a/autoload/codefmt/juliaformatter.vim
+++ b/autoload/codefmt/juliaformatter.vim
@@ -21,7 +21,7 @@ let s:checkedInstall = 0
 " Formatter: JuliaFormatter
 function! codefmt#juliaformatter#GetFormatter() abort
   let l:installer =
-        \ maktaba#path#Join([expand('<sfile>:p:h:h'), 'bin', 'julia', 'install'])
+        \ maktaba#path#Join([s:plugin.location, 'bin', 'julia', 'install'])
   let l:formatter = {
         \ 'name': 'JuliaFormatter', 'setup_instructions': 'Run ' . l:installer}
 
@@ -59,7 +59,8 @@ function! codefmt#juliaformatter#GetFormatter() abort
     endfor
     let l:exec = s:plugin.Flag('julia_format_executable')
     if empty(l:exec)
-      let l:cmd = [expand('<sfile>:h:h/bin/formatjulia.jl')]
+      let l:cmd = [maktaba#path#Join(
+            \ [s:plugin.location, 'bin', 'julia', 'formatjulia.jl'])]
     else
       " Split the command on spaces, unless preceeded by a backslash
       let l:cmd = split(l:exec, '\\\@<! ')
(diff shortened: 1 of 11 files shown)


flwyd:
Added vroom test.

autoload/codefmt/juliaformatter.vim
```diff
@@ -0,0 +1,75 @@
...
+let s:plugin = maktaba#plugin#Get('codefmt')
+
+let s:checkedInstall = 0
+
+""
+" @private
+" Formatter: JuliaFormatter
+function! codefmt#juliaformatter#GetFormatter() abort
+  let l:installer =
+        \ maktaba#path#Join([expand('<sfile>:p:h:h'), 'bin', 'julia', 'install'])
```
dbarnett, line 24: What about `maktaba#path#Join([s:plugin.location, 'bin', 'julia', 'install'])` here, to make it a little more robust about where this particular sfile sits?

dbarnett (approved):
> This "formatter script in the plugin" approach hasn't been taken before in `vim-codefmt`

I think it's fine but do wonder what the steps look like if someone wants to invoke identical formatting outside of vim.

autoload/codefmt/juliaformatter.vim
flwyd: Oh cool, I didn't realize `plugin.location` was a thing.  That's definitely better.

flwyd:
Yeah, I was a little concerned about breaking new ground with a plugin-local formatter script.  I don't _think_ this should result in different formatting results within vim vs. outside of vim.  While the script looks kind of long, most of it is handling line ranges.  The rest is setting up and acting on command line flags.  If formatting decisions are encoded in `.JuliaFormatter.toml` files in the source repository, the results should be the same.

It's worth noting that [the JuliaFormatter.vim plugin](https://github.com/kdheepak/JuliaFormatter.vim) also ships with Julia code which calls the `JuliaFormatter` library rather than executing the script which ships with the library.

[One can also](https://domluna.github.io/JuliaFormatter.jl/stable/) run `JuliaFormatter.format_file("mycode.jl")` from the Julia REPL, and it seems like this kind of thing is encouraged, so I'm guessing that formatting choices are done via config rather than by the executable that invokes the library.  I'm new to Julia, though, so if we get feedback from Julia users that this script approach is awkward, we can pursue another approach.

PR #232, opened 2023-11-24
Fix jsonnetfmt passing junk "-path" argument

### LINKED ISSUE #231: jsonnetfmt formatter broken, always fails with `ERROR: unrecognized argument: -p`
The jsonnetfmt formatter doesn't work at all, fails with an error about passing `-path` as an argument:
```
Error formatting file: ERROR(ShellError): Error running: jsonnetfmt -path somepath.jsonnet
ERROR: unrecognized argument: -p
```

Looks like it was just never working. I can't find any evidence that jsonnetfmt ever supported a `--path` argument. That seems to have just been copypasta from another formatter.

dbarnett:
Fixes #231.


malcolmr (approved):
LGTM, thanks!

PR #233, opened 2023-11-27
Update README.md with missing languages

flwyd:
Also sort lists, document which filetypes support Prettier, and match the AutoFormatBuffer example to the language list.  Formatting changed due to `mdformat` autoformatting.


dbarnett (approved):
Thanks, not sure how I missed this PR but all looks good to me.

Issue #180, opened 2021-04-22
Add LuaFormatter


Camouflager:
**Formatter tool**
LuaFormatter
https://github.com/Koihik/LuaFormatter

**Filetype(s)**
lua

**Additional context**
luaformatterfiveone is unmaintained and outdated. It would be nice to add LuaFormatter

PR #237, opened 2024-07-08
add lua formatter

lamchau:
This PR adds [`stylua`](https://github.com/JohnnyMorganz/StyLua) as an alternative formatter for Lua. 
 - [x] adds `Stylua` to README, corrects missing `)`
 - [x] find/replace `formatterfiveone` with `stylua`
 - [ ] `vroom` has been [archived](https://github.com/google/vroom) and was unable to get it to build on an M3 MBP but took inspiration from `black.vroom`, should work? I could use some help testing this

Partially addresses #180.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ codefmt` if codefmt is installed (and helptags have been generated).
 *   Julia ([JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl))
 *   Kotlin ([ktfmt](https://github.com/facebookincubator/ktfmt))
 *   Lua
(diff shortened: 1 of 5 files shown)


dbarnett:
> I could use some help testing this

Here ya go: https://github.com/dbarnett/vim-codefmt/commit/4384b264f6ac17856ee723fb6fcd11b8b8769099

I'd also suggest you make --stdin-filepath conditional and pass the short name instead of full path, like I did for another formatter here: https://github.com/google/vim-codefmt/blob/e0e113ecfed37309af32b11df8af6fbc17463e0d/autoload/codefmt/ruff.vim#L23-L25. Pretty sure formatters don't need or want the full absolute path.

And vroom is unarchived now FWIW.

PR #239, opened 2024-08-27
ruff formatter: Fix bug running non-stdin version

### LINKED ISSUE #223: Add ruff as formatter for Python
**Formatter tool**
[ruff](https://beta.ruff.rs/docs/)

**Filetype(s)**
- python

**Additional context**
- 10-100x faster 
- Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing tens or hundreds of times faster than any individual tool.

dbarnett:
Forgot to pass the "-" placeholder for stdin. Apparently if you pass --stdin-filename ruff assumes stdin, but if there's no filename it can get confused and format everything in-place instead of formatting the editor buffer.

Fixes #223 (properly).

### DIFF
--- a/autoload/codefmt/ruff.vim
+++ b/autoload/codefmt/ruff.vim
@@ -19,7 +19,7 @@ let s:plugin = maktaba#plugin#Get('codefmt')
 function! s:FormatWithArgs(args) abort
   let l:executable = s:plugin.Flag('ruff_executable')
   let l:lines = getline(1, line('$'))
-  let l:cmd = [l:executable, 'format'] + a:args
+  let l:cmd = [l:executable, 'format'] + a:args + ['-']
   if !empty(@%)
     let l:cmd += ['--stdin-filename=' . @%]
   endif
--- a/vroom/ruff.vroom
+++ b/vroom/ruff.vroom
@@ -21,7 +21,7 @@ system.
   :silent file somefile.py
   % f()
   :FormatCode ruff
-  ! ruff format --stdin-filename=somefile.py.*
+  ! ruff format - --stdin-filename=somefile.py.*
   $ f()
 
 The name or path of the ruff executable can be configured via the
@@ -29,7 +29,7 @@ ruff_executable flag if the default of "ruff" doesn't work.
 
   :Glaive codefmt ruff_executable='/somepath/ruff'
   :FormatCode ruff
-  ! /somepath/ruff format.*
+  ! /somepath/ruff format -.*
   $ f()
   :Glaive codefmt ruff_executable='ruff'
 
@@ -40,7 +40,7 @@ You can format any buffer with ruff specifying the formatter explicitly.
   % if True: pass
 
   :FormatCode ruff
-  ! ruff format.*
+  ! ruff format -.*
   $ if True:
   $     pass
   if True:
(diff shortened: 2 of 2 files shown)


autoload/codefmt/ruff.vim
```diff
@@ -19,7 +19,7 @@ let s:plugin = maktaba#plugin#Get('codefmt')
 function! s:FormatWithArgs(args) abort
   let l:executable = s:plugin.Flag('ruff_executable')
   let l:lines = getline(1, line('$'))
-  let l:cmd = [l:executable, 'format'] + a:args
+  let l:cmd = [l:executable, 'format'] + a:args + ['-']
```
snu5mumr1k, line 22: This should be in the else block, the commands should be:

When the filename is available:

```
ruff format <args> --stdin-filename <filename>
```

When filename is not available

```
ruff format <args> --stdin-filename <placeholder>
```

Where `<placeholder>` can be `-`, or (I would suggest something like `currentbuffer`

vroom/ruff.vroom
```diff
@@ -21,15 +21,15 @@ system.
   :silent file somefile.py
   % f()
   :FormatCode ruff
-  ! ruff format --stdin-filename=somefile.py.*
+  ! ruff format - --stdin-filename=somefile.py.*
```
snu5mumr1k, line 24: Is there a reason to expect something at the end of the line?

snu5mumr1k (changes requested):

autoload/codefmt/ruff.vim
dbarnett: Hmm, there might be a misunderstanding about how the args work. The intended behavior is:
- If there's a known filename: `... <args> - --stdin-filename=foo.py`
- Otherwise: `... <args> -`

Either way, it should have the special `-` placeholder to signal that it should read from stdin, not try to format files in-place, and the `--stdin-filename` value is just an optional hint of the filepath if available to help the formatter understand what it's formatting (for instance, if the project ruff config had per-path overrides).

vroom/ruff.vroom
dbarnett: Yes, the syscall lines from vroom always have some `2>some_outfile` stuff at the end.

autoload/codefmt/ruff.vim
snu5mumr1k: According to the current source, no:
https://github.com/astral-sh/ruff/blob/main/crates/ruff/src/lib.rs#L87-L104

But feel free to leave as is.

snu5mumr1k (approved):

autoload/codefmt/ruff.vim
dbarnett: Isn't that just saying the explicit `-` is optional but not redundant in that case?

If `- --stdin-filename=foo.py` would tend to confuse ruff, then +1 I'd like to change it to drop the `-` in that case, but otherwise I think it expresses the intention better to say we always have a STDIN `-` and sometimes have a `--stdin-filename`.

